PR #5, opened 2023-07-03
docs: clarify library usage requirements

glb:
README now points out that `make quick` will install the module so a person does not need to run that and then also build and install a release.

Also clarifies that the command to build a release is `make release`.


skuenzli (approved):
Good point - thanks for the clarification!

PR #24, opened 2024-11-24
Update cedar dependency to 4.1.0 and add new utility methods

davemssavage:
- Update cedar dependency to 4.1.0 this causes breaking changes in input schema so also upgrading cedarpy version to 1.0.0 as per semantic versioning guidelines
- Add new methods to convert cedar policies to/from cedar natiive and json representation

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cedarpy"
-version = "0.4.1"
-edition = "2021"
+version = "1.0.0"
+edition = "2024"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [lib]
@@ -11,9 +11,9 @@ crate-type = ["cdylib"]
 [dependencies]
 pyo3 = "0.19.0"
 anyhow = "1.0"
-cedar-policy = "~2.2.0"
-cedar-policy-cli = "~2.2.0"
-cedar-policy-formatter = "~2.2.0"
+cedar-policy = "~4.1.0"
+cedar-policy-cli = "~4.1.0"
+cedar-policy-formatter = "~4.1.0"
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = "1.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cedarpy"
 version = "1.0.0"
-edition = "2024"
+edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [lib]
--- a/cedarpy/__init__.py
+++ b/cedarpy/__init__.py
@@ -40,7 +40,7 @@ class AuthzResult:
 
     @property
     def decision(self) -> Decision:
-        return Decision[self._authz_resp['decision']]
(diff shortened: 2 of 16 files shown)


skuenzli:
Thank you David! We really appreciate the PR and will take a look this week.

Cargo.toml
```diff
@@ -11,9 +11,9 @@ crate-type = ["cdylib"]
 [dependencies]
 pyo3 = "0.19.0"
 anyhow = "1.0"
-cedar-policy = "~2.2.0"
-cedar-policy-cli = "~2.2.0"
-cedar-policy-formatter = "~2.2.0"
+cedar-policy = "~4.1.0"
```
davemssavage, line 14: There is a 4.2.2 release available on https://docs.rs/cedar-policy/latest/cedar_policy/ however the public docs https://docs.cedarpolicy.com/ currently refers to 4.1 hence this choice

skuenzli:
David - thank you for putting all this together.  I hadn't anticipated a change to Cedar's schema.  So I need to learn more about that change and think about how to handle it.  Thank you for providing references.

Just an fyi... I will try to take a look at this soon and provide some concrete feedback and get the change on its way to merging.

I'm headed to re:Invent.  I'd love to connect if you'll be there.

davemssavage:
Yeah unfortunately not going to re:Invent, no worries not urgent

Cargo.toml
skuenzli: Thanks for thinking about the versioning.

I think we might jump to a 4.x version number to communicate support for Cedar 4.x

It looks like CedarJavaFFI did that:
https://github.com/cedar-policy/cedar-java/blob/main/CedarJavaFFI/Cargo.toml#L9

Not asking you to do anything here. Just FYI..

cedarpy/__init__.py
```diff
@@ -158,3 +158,24 @@ def format_policies(policies: str,
...
+
+def policies_from_json_str(policies: str) -> str:
+    """Convert a json cedar policy file to a cedar policy file.
+
+    :param policies is a str containing the policies to be converted
+
+    :returns the cedar formatted policy
+    :raises ValueError: if the input policies cannot be parsed
+    """
+    return _internal.policies_from_json_str(policies)
```
skuenzli, line 181: Thanks for adding these functions.

Do you manage your Cedar policies in JSON?

davemssavage: It was to give some sort of parity with the Rust API. In Rust it's possible to parse the policy to the API objects and use this to check what has been generated, e.g. if you want to do some checking on the policy that is not supported by the schema. However returning the Cedar API objects as python objects seemed pretty complex. By allowing cedar policies to be converted to JSON it's probably good enough to be able to see what was generated and do some external validation, the other method to go from json back to cedar native just seemed sensible if you allow from native to json.

davemssavage: Thinking about it's probably worth adding a note to the docstring to warn others of this detail

skuenzli:
Overall, these changes look great to me. I am going to try and mirror Cedar's conventions around branch naming.

So will merge this into a `dev/4.1.x` branch for some tweaking, then `main`, then eventually a `release/4.1.x` for maintenance.

Already branched `main` into `release/2.2.x`.

skuenzli:
Thank you David!

PR #65, opened 2026-05-05
fix: surface invalid schema as NoDecision with diagnostic error (#27)

### LINKED ISSUE #27: cedar-py swallow errors if invalid schema is provided
Hi,

I am planning to integrate Cedar policy engine in our application via cedar-py binding library. I noticed when passing an invalid schema, `cedar-py` will ignore it and only validate the request against entities and policies. There were no indication from `AuthzResult` response neither.

Would it make sense to have the expected behavior to fail the the authorization if a non-empty invalid schema is provided? Happy to raise a PR from my end as well

Thanks!

rupivbluegreen:
## Summary
- Fixes #27: cedar-py used to silently swallow invalid schemas, returning a real Allow/Deny computed without the schema.
- `make_schema` now pushes parse errors into the same `errs` vec used by `make_entities` and policy parsing. The existing `errs.is_empty()` check then routes the response through `make_authz_result_for_errors`, producing `NoDecision` + a diagnostic error — matching the behavior already in place for invalid policies and entities.
- Covers both Cedar schema syntax and JSON schema parse failures.

## Behavior
Before:
```python
authz_result = is_authorized(request, policies, entities, schema="not a schema")
authz_result.decision         # Decision.Allow (or Deny) — schema silently ignored
authz_result.diagnostics.errors  # []
```

After:
```python
authz_result = is_authorized(request, policies, entities, schema="not a schema")
authz_result.decision         # Decision.NoDecision
authz_result.diagnostics.errors  # ["failed to parse schema: ..."]
```

This aligns with the direction @skuenzli suggested in [issue #27](https://github.com/k9securityio/cedar-py/issues/27#issuecomment-...): model the failure as `NoDecision` with diagnostics rather than silent fallback.

## Test plan
- [x] Added two regression tests in `tests/unit/test_authorize.py` (Cedar-syntax and JSON-syntax invalid schemas) — both fail before the fix and pass after.
- [x] `pytest tests/unit` — 53 passed.
- [x] `make integration-tests` — 69 passed, 5 skipped, 0 failed.
- [x] No change in behavior for valid schemas, `None` schemas, or empty/whitespace-only schemas (existing tests cover these).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,7 +136,7 @@ fn is_authorized_batch(requests: Vec<HashMap<String, String>>,
 
     // parse schema
     let t_start_schema = Instant::now();
-    let schema = make_schema(&schema, verbose);
+    let schema = make_schema(&schema, verbose, &mut errs);
     let t_parse_schema_duration = t_start_schema.elapsed();
 
     // load entities
@@ -356,7 +356,7 @@ fn make_entities(entities_str: String, schema: &Option<Schema>, errs: &mut Vec<E
     }
 }
 
-fn make_schema(schema_str: &Option<String>, verbose: bool) -> Option<Schema> {
+fn make_schema(schema_str: &Option<String>, verbose: bool, errs: &mut Vec<Error>) -> Option<Schema> {
     let schema: Option<Schema> = match &schema_str {
         None => None,
         Some(schema_src) => {
@@ -377,6 +377,7 @@ fn make_schema(schema_str: &Option<String>, verbose: bool) -> Option<Schema> {
                         if verbose {
                             println!("!!! could not construct schema from JSON: {}", json_err);
                         }
+                        errs.push(Error::msg(format!("failed to parse schema as JSON: {}", json_err)));
                         None
                     }
                 }
@@ -387,12 +388,13 @@ fn make_schema(schema_str: &Option<String>, verbose: bool) -> Option<Schema> {
                         if verbose {
                             println!("!!! could not construct schema from str: {}", str_err);
                         }
+                        errs.push(Error::msg(format!("failed to parse schema: {}", str_err)));
                         None
                     }
                 }
             }
         }
-    };    
+    };
     schema
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -377,7 +377,7 @@ fn make_schema(schema_str: &Option<String>, verbose: bool, errs: &mut Vec<Error>
                         if verbose {
                             println!("!!! could not construct schema from JSON: {}", json_err);
                         }
-                        errs.push(Error::msg(format!("failed to parse schema as JSON: {}", json_err)));
+                        errs.push(Error::msg(format!("failed to parse schema from JSON: {}", json_err)));
                         None
                     }
                 }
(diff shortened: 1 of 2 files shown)


skuenzli:
@rupivbluegreen - thank you for this PR! 

I reviewed the change quickly and I think it makes sense overall.

My main question is how to introduce this change to `is_authorized`.

1. if it's strictly better, then make it the default behavior 
2. if there are valid use cases for ignoring a broken schema (seems unlikely), then opt-in with an `validate_schema=False` param on `is_authorized`

But I'm having a hard time coming up with valid use cases for ignoring a broken schema.

Consider the first unit tests that you provided:
```
    def test_is_authorized_with_invalid_cedar_schema_returns_no_decision(self):
        # Regression for https://github.com/k9securityio/cedar-py/issues/27
        # Invalid (non-empty) schemas previously returned a real decision while
        # silently dropping the schema. Now they must surface as NoDecision +
        # a diagnostic error.
        policies = self.policies["alice"]
        entities = load_file_as_str("resources/sandbox_b/entities.json")
        invalid_schema = "this is definitely not a cedar schema"

        request = {
            "principal": 'User::"alice"',
            "action": 'Action::"view"',
            "resource": 'Photo::"alice_w2.jpg"',
            "context": json.dumps({"authenticated": False}),
        }

        authz_result: AuthzResult = is_authorized(request, policies, entities, schema=invalid_schema)
        self.assertEqual(Decision.NoDecision, authz_result.decision)
        self.assertEqual(1, len(authz_result.diagnostics.errors))
        self.assertIn("schema", authz_result.diagnostics.errors[0].lower())
```

Alice has this policy:

```
permit(
    principal == User::"alice",
    action == Action::"view",
    resource
)
;
```

This (relevant) entity:

```
  {
    "uid" : {
      "type" : "Photo",
      "id" : "alice_w2.jpg"
    },
    "attrs": {
      "private": true,
      "account": { "__entity": { "type" : "Account", "id" : "alice" } },
      "admins": []
    },
    "parents": [
      {
        "__entity": { "type" : "Account", "id" : "alice" }
      }
    ]
  }
```

Currently cedarpy (4.8.1) `is_authorized` evaluates to `Decision.Allow`. cedarpy+cedar effectively ignore the request context because there is no schema that says, "when evaluating the view action, verify the Account is authenticated".

skuenzli:
After additional consideration, I think requiring a valid schema if the caller provides one is the right choice.

I just want to make one minor clarification on the schema error messages

skuenzli:
Hey @rupivbluegreen — pushed a small follow-up (`f84becc`) on top of your fix: just tightened the two schema parse error messages from `"failed to parse schema: …"` / `"failed to parse schema as JSON: …"` to `"failed to parse schema from Cedar: …"` / `"failed to parse schema from JSON: …"` so the wording is symmetric and tells the caller which format the parser tried. No behavior change; your tests still pass as written. Thanks for the fix — the diagnostic-error path is exactly what we wanted for #27.

rupivbluegreen:
Clear, makes sense to tighten it.

skuenzli:
Thank you for the contribution!

Issue #29, opened 2025-06-17
Support for policy templates


jgmills98:
Hi,

I am currently implementing a poc with cedar using this library. Is there any support for passing in template-linked policies?

Thank you for your time and work on this project!

skuenzli:
Hi Jordan - the library doesn't support template-linked policies right now. 

But if you want to collaborate on adding that support, we'd be open to it.

(you're welcome!)

Iamrodos:
Hi @skuenzli, I'd like to pick up template-linked policy support from this issue. I've been through the crate API and the current binding, and the gap is smaller than I expected, so I want to agree the Python-facing surface here before opening a PR.

**What's already there**

`PolicySet.from_str` already parses templates (the `?principal` / `?resource` form) into the set. They just sit inert: they aren't counted in `__len__` and they authorize nothing, because nothing is linked to them. So the missing piece is linking a template to concrete entities, not parsing.

**Proposed Python surface**

Keep the immutable handle pattern from the reusable PolicySet work. Linking returns a new PolicySet rather than mutating in place:

```python
base = PolicySet.from_str(template_text)

linked = base.link(
    template_id="share-album",
    new_id="share-album-for-jane",
    values={"?principal": 'User::"jane"', "?resource": 'Album::"vacation"'},
)

result = is_authorized(request, linked, entities)
```

Slot values would accept the same two forms the request fields already accept: either Cedar syntax (`User::"jane"`) or the `{"type": ..., "id": ...}` JSON form.

This stays additive and backward compatible. Existing `is_authorized(request, policies, entities)` callers are untouched, and a linked PolicySet is still just a PolicySet.

**Decisions, with where I'm leaning**

1. Scope of v1. I'd include the read side from the start: listing templates, listing a template's links, and `unlink`, alongside linking itself. It maps directly onto crate methods that are already there (`templates()`, `get_linked_policies()`, `unlink()`), so I don't think it's much extra code or interface to get right up front, and it makes the feature feel complete rather than write-only. If you'd rather nail the basics first, I'm happy to cut v1 down to linking only and follow up with introspection, but my lean is to do it in one go.

[...]

PR #66, opened 2026-05-05
feat: honor @id("...") annotation as the policy id

rupivbluegreen:
## Summary
Addresses item 2 of the feature wish-list in #29: cedar parses `@id("...")` annotations on policies and templates, but cedarpy used the auto-generated `policy0`/`policy1`/... ids regardless. This made `AuthzResult.diagnostics.reasons` and validation errors reference opaque numeric ids instead of the names users wrote in their policies.

After this PR:

```cedar
@id("alice_can_view")
permit(principal == User::"alice", action == Action::"view", resource);
```

```python
authz_result = is_authorized(req, policies, entities)
authz_result.diagnostics.reasons   # ["alice_can_view"]   (was ["policy0"])
```

Same applies to `validate_policies`: `ValidationError.policy_id` now reports the `@id`-named id when validation fails.

## Implementation
- New helper `rename_from_id_annotation` in `src/lib.rs`, modeled on the same-named function in `cedar-policy-cli`.
- Applied right after `PolicySet::from_str` in both `is_authorized_batch` and `validate_policies`.
- Two failure modes are surfaced as diagnostics rather than panicking (the cedar-policy-cli reference uses `.expect(...)` which would panic in a library context):
  - **Duplicate id** — two policies (or templates) resolving to the same `@id` produce `Decision.NoDecision` with `"duplicate policy id after applying @id annotation"` in `diagnostics.errors`.
  - **Malformed PolicyId** — `PolicyId::from_str` is currently infallible upstream, but is defensively handled in case it grows fallibility.

Templates are also handled, even though cedarpy doesn't surface a templates API yet (#29 main item) — keeping the helper symmetric with cedar-policy-cli avoids re-touching this code when template support lands.

## Test plan
- [x] Added 4 unit tests:
  - `@id` rename surfaces in `diagnostics.reasons` for the matched policy
  - mixed annotated/un-annotated policies — annotated keeps custom id, un-annotated retains `policyN`
  - duplicate `@id` returns `NoDecision` + diagnostic
  - `validate_policies` reports the `@id`-named id in `ValidationError.policy_id`
- [x] `pytest tests/unit` — 55 passed, 0 failed.
- [x] `make integration-tests` — 69 passed, 5 skipped, 0 failed.
- [x] No change in behavior for policies without `@id` — existing tests asserting `policy0`/`policy1`/... continue to pass.

## Notes
- Independent of #65 (no shared lines).
- Worth flagging: this is a small **behavior change** — anyone authoring policies that happen to include an `@id("name")` annotation today (maybe as documentation) would suddenly see that name in their reasons/error output. The Cedar engine treats `@id` as having this exact semantic meaning, so the change aligns cedarpy with cedar-policy-cli — but if you'd prefer this be opt-in via a flag, happy to add one.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -123,7 +123,14 @@ fn is_authorized_batch(requests: Vec<HashMap<String, String>>,
     // parse policies
     let t_parse_policies = Instant::now();
     let policy_set = match PolicySet::from_str(&policies) {
-        Ok(pset) => pset,
+        Ok(pset) => match rename_from_id_annotation(pset) {
+            Ok(renamed) => renamed,
+            Err(e) => {
+                println!("{:#}", e);
+                errs.push(e);
+                PolicySet::new()
+            }
+        },
         Err(parse_errors) => {
             let err_message = format!("policy parse errors:\n{:#}",
                                       parse_errors.to_string());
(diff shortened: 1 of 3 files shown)


skuenzli:
@rupivbluegreen - thank you for this PR. This is a nice ergonomic improvement. 

Today is packed. I plan to review it (my) tomorrow morning.

skuenzli:
Pushed two follow-ups (`65eb6e9`, `b1be887`):

**`65eb6e9` — test: lock in `validate_policies` duplicate-@id behavior**
Added a regression test capturing the exact diagnostic shape returned when two policies share an `@id`. The motivation was twofold: (1) the change in `is_authorized` for duplicate `@id` is already covered by your `test_id_annotation_duplicate_returns_no_decision`, but the parallel case in `validate_policies` had no test; and (2) since this is a behavior change vs. pre-PR-66 (where duplicates were inert), having an explicit test makes any future drift visible in diffs.

**`b1be887` — fix: surface original policy id when `@id` rename fails**
The `validate_policies` rename-error path was setting `ValidationError.policy_id = ""`, which doesn't help callers locate the offending policy in their input. Changed `rename_from_id_annotation` to return `Result<PolicySet, (String, Error)>` where the `String` is the offending policy's pre-rename id (e.g. `"policy1"` for the second of two duplicates), and propagated it through `validate_policies`. Updated the duplicate-`@id` test from `65eb6e9` to assert the new value, so the change is visible in the diff.

`is_authorized_batch` discards the id (its diagnostic surface is a flat list of error strings), so behavior there is unchanged — only the helper's return type moved.

Reasoning for splitting the test commit from the fix: the test commit captures *current* behavior; the fix commit then changes it.

Also merged latest `main` to resolve conflicts with PR #65 (the schema fix that landed yesterday) — `tests/unit/test_authorize.py` had overlapping test additions; both sets are preserved.

skuenzli:
Thank you for the contribution!

skuenzli:
For reference: the implementation approach landed here has been replaced by **Path B** in #75. cedar-py now resolves `@id` annotations at response-serialization time rather than by renaming `PolicySet` ids at parse time.

The user-visible behavior introduced by this PR is preserved (`@id` value surfacing in `AuthzResult.diagnostics.reasons` and `ValidationError.policy_id`); the change is internal. The motivation was twofold:

- **Correctness**: per the [Cedar annotation docs](https://docs.cedarpolicy.com/policies/syntax-policy.html#term-parc-annotations), annotations have no impact on policy evaluation and `@id` is not special in the Cedar language. cedar-py is a Cedar API consumer, not a CLI, and handling `@id` at the response surface is the more faithful approach.
- **Performance**: the per-call `PolicySet` rebuild caused the regression tracked in #68. Path B is `O(matched policies)` per response instead of `O(|PolicySet|)` per call.

Context and design discussion: #74. Behavior change vs this PR: duplicate `@id` values are now accepted (the NoDecision / `validation_passed=False` behavior introduced here was a Path A artifact, not a Cedar-engine semantic).

PR #70, opened 2026-05-09
fix: make @id annotation resolution opt-in via resolve_policy_ids_from_annotations

skuenzli:
## Summary

Resolves #68. Makes `@id` annotation resolution opt-in via a new `resolve_policy_ids_from_annotations` parameter on `is_authorized`, `is_authorized_batch`, and `validate_policies`. The parameter defaults to `False`, restoring v4.8.0 performance on the default code path.

Also fixes a separate latent issue with the `make release` Makefile target (was producing debug wheels despite the name), and updates `CLAUDE.md` with cedar-API gotchas and benchmark-process lessons learned during the investigation.

## Commits

1. **`fix: make @id annotation resolution opt-in via resolve_policy_ids_from_annotations`** — the perf fix proper
2. **`docs: capture cedar API gotchas and benchmark process lessons in CLAUDE.md`** — guidance for future maintainers / AI assistants
3. **`fix: build release-mode wheel in 'make release' target`** — local Makefile bug fix; PyPI publishing was already correct

## Why

PR #66 introduced an unconditional `PolicySet` rebuild on every `is_authorized*` / `validate_policies` call to honor `@id` annotations. The rebuild cost ~15% on complex-policy workloads in release mode (~30% in debug), even when no policy used `@id` at all — because cedar's `Policy::clone()` and `PolicySet::add()` do real work proportional to `|policies|`.

This was caught when preparing the v4.8.2 release: `make benchmark-compare` failed loudly on multiple workloads. A real-world downstream test (running cedarpy in a dependent product) confirmed the production-mode regression.

## Performance

N=5 release-mode median, vs v4.8.0 baseline (microseconds):

| Benchmark | v4.8.0 | PR #66 (always rename) | This PR (default off) |
|---|---|---|---|
| `test_complex_policy` | 286 | 329 (**+14.9%**) | **275 (-3.8%)** |
| `test_simple_policy_deny` | 152 | 153 (+1.1%) | **148 (-2.5%)** |
| `test_medium_policy` | 216 | 213 (-1.0%) | **191 (-11.2%)** |
| `test_small_entity_set` | 188 | 213 (+13.3%) | **190 (+1.1%)** |
| `test_batch_complex_policy` | 1234 | 1332 (+8.0%) | **1242 (+0.6%)** |

Default-off lands within v4.8.0's noise floor on every benchmark. Per `out/v4.8.2-bisect/summary.md` (full bisect data, 8 states × 5 runs × 2 build modes).

## API change

```python
[...]

Issue #69, opened 2026-05-09
benchmark process improvements: release builds, refreshed baseline, multi-run signal


skuenzli:
## Goal

Improve the perf benchmark process so that future regressions like the one in #68 are caught reliably and with less manual investigation. Three concrete improvements, summarized below.

## 1. Switch benchmarks to release builds

`Makefile` currently runs `maturin develop` (debug mode) for `benchmark`, `benchmark-save`, and `benchmark-compare`. Debug mode amplifies the cost of allocations and bounds-checks, producing run-to-run variance large enough that single-run regressions of ~30% are routinely just noise. The N=5 bisect on #68 confirmed this: debug mode showed `test_complex_policy` regressing +30%, while release mode (the only mode users actually run) showed +14.9%.

Switching all three Makefile targets to `maturin develop --release` does two things:
- **Variance drops dramatically.** Release-mode means stay within a few % across runs vs debug-mode's swings of 20%+ on tail outliers (compare release vs debug columns in `out/v4.8.2-bisect/summary.md`).
- **Numbers reflect production.** Whatever a regression test reports is what users actually experience.

Acceptance: all three benchmark targets in `Makefile` use `maturin develop --release`.

## 2. Update the baseline

The current `tests/benchmark/results/baseline.json` was recorded at v4.8.0 in **debug** mode. Switching to release in goal 1 above invalidates the comparison entirely (release-mode runs are ~8x faster than debug-mode baseline, so every comparison would trivially "pass").

After goal 1 lands, regenerate `baseline.json` from a release-mode build at the current `main` tip (post-fix-for-#68). This becomes the new comparison reference.

Recommend:
- Use `make benchmark-save` to record the new baseline (after the Makefile change above)
- Replace `tests/benchmark/results/baseline.json`
- Note the build mode and source tip in the commit message that updates the baseline

## 3. Make `make benchmark-compare` a trustworthy multi-run signal

The N=5 multi-run aggregation we built for the #68 investigation was **the** thing that turned the perf data from misleading to actionable. Single-run variance was so high that the same code ran twice could produce regression deltas off by 50 percentage points. We should make multi-run-with-aggregation the default path, not a one-off script we wrote during a fire.

Two routes worth investigating, in this order:

### 3a. See if `pytest-benchmark` solves this natively

Check whether `pytest-benchmark` supports any of:
- A multi-invocation mode (run the same benchmarks N times across separate pytest sessions and aggregate)
[...]

PR #73, opened 2026-05-10
chore(benchmark): record 6th historical state for PR #70 (restores v4.8.0 perf)

skuenzli:
## Summary

Adds a 6th state to the historical benchmark record: the tip of `fix/optional-resolve-policy-ids` (1511625, the branch backing PR #70). The data lets the historical record verify whether PR #70's opt-in approach actually restores v4.8.0 performance on the default code path.

## Headline finding for the PR #70 merge decision

**PR #70 restores v4.8.0 performance on the default path.** Δ medians vs v4.8.0:

| Benchmark | main HEAD | PR #70 tip |
|---|---|---|
| `test_complex_policy` | **+14.9%** | **−0.3%** |
| `test_medium_policy` | +7.5% | +0.1% |
| `test_small_entity_set` | +7.4% | −1.8% |
| `test_simple_policy_allow` | +1.7% | −1.0% |
| `test_batch_complex_policy` | +2.3% | −0.8% |

All Δs at PR #70 are within ±2% of v4.8.0 — i.e., within the run-to-run noise we observed during the N=7 backfill experiment in PR #71. The design intent of #70 is confirmed: when `resolve_policy_ids_from_annotations` defaults to `False`, the per-call `PolicySet` rebuild that PR #66 introduced is skipped, and timings match v4.8.0.

## What this PR adds

- New STATES entry `pr70_1511625:1511625:PR70 opt-in @id annotation resolution` in `tests/benchmark/capture_history.sh`, with the matching `states-manifest.json` record.
- 5 captured release-mode native pytest-benchmark JSONs at `tests/benchmark/results/Darwin-CPython-3.11-64bit/000{3..7}_pr70_1511625-runN.json`.
- Per-commit summary `tests/benchmark/results/history/1511625.json`.
- Regenerated `tests/benchmark/results/HISTORY.md` with the new row in each per-benchmark table.

## Implementation notes

- Captured against the same `tests/benchmark/test_benchmark_authorize.py` blob as the other states (verified via `git ls-tree`), so the comparison is apples-to-apples.
- The other 5 per-commit summary JSONs (`v4.8.0.json` etc.) are also touched: only their `captured_at` timestamps changed when the aggregator re-ran. Stats values are unchanged.
- This PR does not measure the **opt-in** code path (`resolve_policy_ids_from_annotations=True`). That measurement is a separate add: a benchmark variant calling the API with the parameter set, captured the same way. Out of scope here; tracked as future work in #69.

## Test plan

- [ ] Read `tests/benchmark/results/HISTORY.md` — verify the new row `PR70 opt-in @id annotation resolution` appears in each table.
- [ ] For `test_complex_policy`, confirm the Δ median for PR #70 is near 0% (within noise) — i.e., perf is restored to v4.8.0 levels.
- [ ] `tests/benchmark/results/history/1511625.json` exists, has `n_runs: 5` per benchmark, and the `commit.label` reads "PR70 opt-in @id annotation resolution".
- [ ] `tests/benchmark/results/history/states-manifest.json` lists the new state.


skuenzli:
Mooted by #75

PR #75, opened 2026-05-11
feat: resolve @id annotations at response time (Path B)

### LINKED ISSUE #68: perf: rename_from_id_annotation regresses is_authorized by up to 31% on policy-heavy workloads
## Problem

Running `make benchmark-compare` against the v4.8.0 baseline (`tests/benchmark/results/baseline.json`) on `main` (`ca5d83c`) fails the **mean ≥ 15%** regression threshold on **7 of 26 benchmarks**. Worst case: `test_complex_policy` at **+31.7%**.

Caught while preparing the v4.8.2 release; the release is currently paused until this is resolved.

## Failing benchmarks (mean regression > 15%)

| Benchmark | Baseline mean (ms) | Now mean (ms) | Mean regression |
|---|---|---|---|
| `test_complex_policy` | 2.5692 | 3.3847 | **+31.7%** |
| `test_context_as_json_string` | 4.5216 | 5.9017 | +30.5% |
| `test_entities_as_list` | 4.5020 | 5.6712 | +26.0% |
| `test_entities_as_json_string` | 4.3879 | 5.3460 | +21.8% |
| `test_medium_policy` | 1.7851 | 2.1381 | +19.8% |
| `test_small_entity_set` | 1.8256 | 2.1866 | +19.8% |
| `test_sandbox_b_batch_mixed_actions` | 13.6846 | 16.0448 | +17.2% |

`test_simple_policy_*` regresses ~10% (under threshold). The pattern: regression scales with the number of policies in the input, not with anything about the request.

## Isolation

The baseline was last refreshed at v4.8.0. Two PRs have landed since:

- #65 — invalid schema → `NoDecision`. Adds one `errs.push` in `make_schema`'s error path. Should not affect success-path benchmarks (none of the benchmarks trigger schema parse errors).
- #66 — `@id` annotation honored as policy id. Adds `rename_from_id_annotation`, called unconditionally on every `is_authorized` / `is_authorized_batch` / `validate_policies` call.

The benchmarks in our suite do not use `@id` annotations, but they still pay the full cost of `rename_from_id_annotation`. The helper:

1. Constructs a fresh `PolicySet::new()`
2. Iterates every template — clones each (`Template::clone()`) and re-adds via `add_template`
3. Iterates every policy — clones each (`Policy::clone()`) and re-adds via `add`

The work is proportional to `|policies|`, which matches the observed scaling: trivial policy sets regress ~10%, complex policy sets regress ~32%. The per-policy cost is not free even though cedar's AST template is Arc-shared internally ([`ast::Policy`](https://docs.rs/cedar-policy-core/4.8.2/cedar_policy_core/ast/struct.Policy.html) holds an `Arc<Template>`): each `Policy::clone()` still clones the `LosslessPolicy` representation cedar keeps alongside the AST for round-tripping, and each [`PolicySet::add`](https://docs.rs/cedar-policy/4.8.2/cedar_policy/struct.PolicySet.html#method.add) performs an additional `ast.clone()`, an internal `ast::PolicySet::add` (validation + indexing), and a `HashMap` insert.

`rename_from_id_annotation` was modeled on cedar-policy-cli's same-named function. The CLI runs the rename once per process (parse → authorize → exit), so its cost is amortized over a single invocation. cedarpy's lifecycle is different — the same policies get evaluated potentially thousands of times per process, and the rename runs on every call.

## Proposed solutions

We're exploring a few options, including:

- **Short-circuit the rebuild** when no policy or template has an `@id` annotation — small, low-risk fix that eliminates the regression for the common case where `@id` is unused, while keeping current behavior for callers who do use the feature.
- **Lazy id translation at result-construction time** — keep the `PolicySet` produced by `from_str` unchanged; only resolve `@id` annotations when constructing `AuthzResult.diagnostics.reasons` or `ValidationError.policy_id`. Eliminates the rebuild cost entirely, including for `@id`-using callers.
- **Make policy renaming an opt-in parameter on `is_authorized` / `is_authorized_batch` / `validate_policies`** (e.g. `honor_id_annotations: bool`, default `False`). Callers opt into the `@id` rename with documented, explicit behavior; callers who don't pay nothing. Trades zero default overhead for a small API surface addition.

## Acceptance criteria

- [ ] All 26 benchmarks within the configured thresholds (`median:5%`, `mean:15%`) on a fresh `make benchmark-compare` run
- [ ] Unit and integration tests still pass

### LINKED ISSUE #74: Define and implement support for `@id` annotations
## Background

The Cedar `@id("...")` annotation is *not* part of Cedar's policy-evaluation semantics. Per the [Cedar policy syntax docs](https://docs.cedarpolicy.com/policies/syntax-policy.html#term-parc-annotations):

> An annotation has no impact on policy evaluation. … `@id` is not special in the Cedar language, it behaves like any other annotation. The Cedar CLI uses the `@id` annotation to set policy IDs, but other interfaces, such as the Cedar APIs, have other ways to set policy IDs.

Confirmed against the vendored Cedar source (`third_party/cedar/`):

- `cedar-policy-core/src/evaluator.rs` and `authorizer.rs` do not read user-provided annotations during evaluation. `@id` cannot influence whether a policy applies to a request.
- `cedar-policy-cli/src/lib.rs:660–676` defines `rename_from_id_annotation` and calls it once at policy-load time inside `read_policy_set` (line 725). This is a CLI labeling convention applied before any template instancing.
- `cedar-policy/src/api.rs:1260–1265` exposes `PolicySet::annotation(id, key)` so library consumers can look up `@id` (or any other annotation) without rebuilding the `PolicySet`.

## Current state in cedarpy

cedarpy currently implements **Path A** — rename the `PolicySet` so the renamed ids appear natively in `Diagnostics.reason` and `ValidationError.policy_id`:

- Introduced in #66, applied unconditionally on every `is_authorized*` and `validate_policies` call.
- Caused the perf regression tracked in #68 (~+15% on `test_complex_policy`, paid per-call even when no policy uses `@id`).
- Made opt-in via `resolve_policy_ids_from_annotations=False` default in #70 (currently OPEN). The opt-in restored v4.8.0 perf on the default path but still pays full rebuild cost when users want `@id` to surface.

PR #66's framing ("the Cedar engine treats `@id` as having this exact semantic meaning") was inaccurate. The Cedar engine does not treat `@id` specially; the Cedar CLI does. cedarpy is closer in role to a Cedar API consumer than to the CLI, and the docs explicitly carve out: "other interfaces, such as the Cedar APIs, have other ways to set policy IDs."

## Goal

Provide the developer/operator ergonomics that motivated #66 — human-readable ids in `AuthzResult.diagnostics.reasons` and in `ValidationError.policy_id` when the author writes `@id("...")` — without:

- restricting what users can do with the underlying Cedar engine, or
- imposing per-call rebuild cost on workloads that don't use `@id`.

## Proposed approach: Path B (post-process)

Leave parser-generated `PolicyId`s (`policy0`, `policy1`, …) intact through evaluation. Resolve `@id` only when serializing the `AuthzResponse` / `ValidationResult`:

```rust
for pid in response.diagnostics().reason() {
    let display_id = policy_set.annotation(pid, "id").unwrap_or(pid.as_ref());
    // emit display_id in the JSON response
}
```

Cost: one annotation lookup per *matched* policy in a response (typically a small handful), not `O(|PolicySet|)` per call. Evaluation is untouched. Same idea for `validate_policies`: resolve each `ValidationError.policy_id` before serializing.

## Open questions / scope

- **Response shape.** Replace the `PolicyId` string in-place (smallest API surface, matches Path A behavior) vs. surface both (`policy_id` parser-generated, `display_id` annotation-resolved) so tooling can see ground truth.
- **Default behavior.** Path B's cost is small enough that always-on may be fine. The `resolve_policy_ids_from_annotations` parameter introduced in #70 may no longer be needed; worth re-benchmarking once Path B lands.
- **Duplicate `@id` values.** Path A treats this as `NoDecision` (can't `add` two policies with the same id). Path B can emit the same `display_id` for each matched policy and let callers see the collision in their diagnostics — not necessarily an error.
- **Templates.** What to do once cedarpy surfaces a templates API (#29).
- **Removal of Path A.** `rename_from_id_annotation` in `src/lib.rs` and the `resolve_policy_ids_from_annotations` parameter from #70 are candidates for removal/replacement once Path B covers the same ergonomics.

## Related

- #29 — template support (original wishlist; `@id` support was item 2)
- #66 — Path A implementation (MERGED)
- #68 — perf regression caused by Path A
- #70 — opt-in workaround (OPEN); likely superseded by this work
- #69 — benchmark process improvements (provides the data infrastructure to measure Path B accurately)

skuenzli:
## Summary

Resolves #74. Replaces the unconditional `PolicySet` rebuild from #66 with a post-process at response-serialization time: parser-generated `PolicyId`s stay intact through evaluation, and `@id` is resolved only when populating `AuthzResult.diagnostics.reasons` and `ValidationError.policy_id`.

Per [Cedar's docs](https://docs.cedarpolicy.com/policies/syntax-policy.html#term-parc-annotations), annotations have no impact on policy evaluation and `@id` is not special in the Cedar language — it's a Cedar CLI labeling convention. cedar-py is a Cedar API consumer, not a CLI, so handling `@id` at the response surface is the more faithful and cheaper approach.

Also fixes #68 at its root: the per-call rebuild is gone entirely. Supersedes #70 (the opt-in workaround); close #70 without merging once this lands.

## Behavior changes vs `main` (still unreleased)

| Surface | Before this PR (post-#66) | After this PR |
|---|---|---|
| `AuthzResult.diagnostics.reasons` on `@id`-annotated policies | `@id` value | `@id` value (unchanged) |
| `ValidationError.policy_id` on `@id`-annotated policies | `@id` value | `@id` value (unchanged) |
| Duplicate `@id` values | `Decision.NoDecision` / `validation_passed=False` with `"duplicate policy id"` diagnostic | Accepted; evaluated independently per Cedar engine semantics |
| `@id("")` and value-less `@id` | Empty-string display id | Falls back to parser-generated id (cedar-py-specific labeling choice) |
| Per-call cost on workloads without `@id` | ~+15% on `test_complex_policy` (rebuild) | Within v4.8.0 noise floor |

No shipped users are affected — #66 has only been on `main` (`[Unreleased]`), never on PyPI.

## Performance

`make benchmark-compare` vs the v4.8.0 release-mode median baseline (medians across multiple runs):

| Benchmark | Baseline median (μs) | Path B median (μs) | Δ median |
|---|---|---|---|
| `test_complex_policy` | 273.3 | 272.4–276.2 | -0.4% to +1.1% |
| `test_batch_simple_policy` | 384.5 | 387.2–388.3 | +0.7% to +1.0% |
| `test_sandbox_b_batch_multiple_users` | 683.1 | 684.4–686.6 | +0.2% to +0.5% |
| `test_batch_complex_policy` | 1234.2 | 1240.5–1253.2 | +0.5% to +1.5% |

All within v4.8.0 noise floor (≤3% on median). Single-run mean-based failures are expected per #69 goal 3 and the CLAUDE.md note on mean noise.

**Implementation gotcha worth flagging for review:** `PolicySet::annotation(pid, "id")` parses `"id"` as a Cedar `Id` on every call. A first draft using it showed a reproducible +9% median regression on `test_batch_simple_policy`. The committed `resolve_display_id` uses `policy.annotations().find(|(k, _)| *k == "id")` instead, which works on raw `&str` keys and is essentially free. Captured in CLAUDE.md.

## API surface

No new parameters. `is_authorized`, `is_authorized_batch`, and `validate_policies` keep their existing signatures. The `resolve_policy_ids_from_annotations` parameter proposed in #70 is **not** added — Path B's cost is small enough that always-on is the right default, and no opt-in is needed.

## Tests

- `pytest tests/unit` — 62 passed
- `make integration-tests` — 69 passed, 5 skipped
- `make benchmark-compare` — medians within noise floor as above

New / rewritten `@id` tests (in `tests/unit/test_authorize.py` and `tests/unit/test_validate.py`):

- `test_id_annotation_duplicates_are_allowed` — duplicates accepted, decision is `Allow`, reasons dedupe to single entry
- `test_id_annotation_does_not_affect_evaluation` — guard for the docs claim that annotations don't affect evaluation; covers alice/bob/carol with two same-`@id` policies and asserts each principal routes correctly
- `test_id_annotation_coexists_with_other_annotations` — stacks `@advice` + `@id` + `@shadow_mode` from the docs' example
- `test_id_annotation_empty_value_falls_back_to_parser_id` — covers both `@id` (no parens) and `@id("")` via `subTest`
- `test_validation_error_reports_parser_id_when_no_id_annotation` — un-annotated policy → `policy0` in `ValidationError.policy_id`
- `test_validate_with_duplicate_policy_id_annotations` — rewritten: now asserts duplicates pass validation

Plus a documented audit of the unit suite against the Cedar annotations docs (1 misalignment resolved, 3 gaps closed, 3 gaps deferred with rationale) recorded in `docs/tasks/74-id-annotation-support.md`.

## Follow-ups

- Close #68 (root cause fixed) and #70 (superseded) once this lands.
- The `feat/issue-69-state-pr70` branch / PR #73 captures the `fix/optional-resolve-policy-ids` benchmark state; both can be closed without merging once this PR lands.
- General-purpose annotations API for non-`@id` keys remains a deferred follow-up; open a separate issue if there's demand.

## Test plan

- [x] CI green
- [x] One round of human review on the `resolve_display_id` implementation choice (raw `&str` keys vs `PolicySet::annotation`)
- [x] Confirm the empty-`@id` fallback behavior (delib choice, diverges slightly from cedar-policy-cli)


skuenzli:
@rupivbluegreen - I have adjusted the approach to supporting `@id` annotations to what is described in the PR after much research and testing, especially for performance.

The key factor in the change in approach is that the [Cedar Policy annotations documentation](https://docs.cedarpolicy.com/policies/syntax-policy.html#term-parc-annotations) says (emphasis added):

> You can attach arbitrary key-value pairs to Cedar policies in the form of annotations. _An annotation has no impact on policy evaluation. Even though annotations are not processed by Cedar_, they are stored as part of the policy and are available for use by services and applications that read and process Cedar policies.

So the role of annotations has been reoriented to support development & user ergonomics. And by contrast, it is _incorrect_ to use `@id` annotations to enforce, e.g. uniqueness. It is likely that cedar-py has consumers today that have 'duplicate' `@id` annotations in their policies and we can't just break them. The Cedar Policy engine allows those policies per specification so cedar-py should just pass those through.

The cedar-policy-cli is closer to the end user than this library, so it is in the position to make choices that this library cannot.

Please let me know what issues you see with this PR, if any.

PR #78, opened 2026-05-13
fix(#77): surface parser policy ids in diagnostics + add id_annotations map

### LINKED ISSUE #77: v4.8.2: diagnostics.reasons collapses tenant identity when two policies share an @id annotation
First — thank you for the work in #74 / #75 to surface `@id` annotations in diagnostics. The readability win is real, and the release notes call the change out clearly. This issue is about a downstream consequence that I think is worth a second look.

## Summary

In v4.8.2, when two distinct policies share the same `@id("...")` annotation, `AuthzResult.diagnostics.reasons` returns the duplicated annotation string instead of the unique parser-assigned `policyN` id. Callers can no longer tell which of the colliding policies determined the decision.

This pattern is the multi-tenant baseline-permit scenario: every tenant authors their copy of a templated `@id("baseline-permit-all")` permit with a tenant-scoped `resource in App::"…"`. In v4.8.1 the diagnostic identified the matching policy as `policy0` (or whatever positional id), letting the caller map back to *which tenant's* policy fired. In v4.8.2 the diagnostic identifies it as `baseline-permit-all` — true for both tenants — and the caller can't disambiguate.

## Minimal reproducer

\`\`\`python
import cedarpy

POLICIES = '''
@id(\"baseline-permit-all\")
permit(principal, action, resource in App::\"app-A\");

@id(\"baseline-permit-all\")
permit(principal, action, resource in App::\"app-B\");
'''

request = {
    \"principal\": 'User::\"alice\"',
    \"action\": 'Action::\"read\"',
    \"resource\": 'App::\"app-A\"',
    \"context\": {},
}

result = cedarpy.is_authorized(request=request, policies=POLICIES, entities=[])
print(\"decision:\", result.decision)
print(\"reasons :\", list(result.diagnostics.reasons))
\`\`\`

Output:

| Version | reasons |
|---|---|
| 4.8.1 | `['policy0']` ← uniquely identifies the App-A policy |
| 4.8.2 | `['baseline-permit-all']` ← indistinguishable from the App-B policy |

Both runs return `Decision.Allow`. The decision itself is correct; only the diagnostic identity is collapsed.

## Why this matters in practice

We're hitting this in [highflame-policy](https://github.com/highflame-ai/highflame-policy) where `PolicyEngine.load_policies_with_ids` exists specifically to give multi-tenant callers a stable caller-supplied id per policy. That layer builds a `{positional_id → caller_id}` map at load time and looks the diagnostic up against it at evaluate time — which works on 4.8.1 and silently misroutes on 4.8.2 (the keys it inserted are `policy0`/`policy1`, but the diagnostic now reports `baseline-permit-all`).

Per the release notes, this is described as a labeling step, not a `PolicyId` rename. But for callers downstream that's a distinction without a difference — the diagnostic surface IS the lookup key. The closed-discussion in #74 makes the case clearly: \`@id\` is inert in Cedar evaluation and is a CLI/UX convention. Multi-tenant deployments break that convention by design: tenants share template text including the annotation but need distinct *evaluation* identities.

## Suggestions

A few directions, ranked by my guess at preference:

1. **Make the new behavior opt-in** via a kwarg, e.g. `is_authorized(..., resolve_policy_ids_from_annotations=False)` matching the constructor-level flag pattern from #70 (closed). Default could be either way; the multi-tenant case just needs *some* way to opt out.
2. **Skip the rename when the annotation is non-unique within the loaded set.** Empty `@id(\"\")` already falls back to the parser-generated id per #75; a similar fallback for duplicate `@id` would preserve readability for the single-tenant happy path while keeping tenants distinguishable when they collide.
3. **Yank/republish as 4.9.0 + revert 4.8.2.** This is a behavior change to a diagnostic surface that callers were relying on (we were), shipped in a patch release. Strict semver would call it minor/major; signaling it as such would have given us time to adapt before the upgrade.
4. Even without (3), a short callout in the release notes that this changes the *content* of `reasons` (not just adds context) would have made it findable when CI broke. The current note frames it as additive.

I'm happy to send a PR for (1) or (2) if either direction sounds right — just let me know which you'd accept and I'll take a swing.

## Environment

- cedarpy 4.8.1 vs 4.8.2
- Python 3.14.4 (Linux, CI) and Python 3.12.7 (macOS, local) — reproduces identically on both

skuenzli:
## Summary

- Reverts the 4.8.2 behavior of renaming `AuthzResult.diagnostics.reasons` entries and `ValidationError.policy_id` to the `@id` annotation value, which collapsed identity when multiple policies shared the same `@id` (e.g. the multi-tenant baseline-permit pattern reported in #77). Both surfaces are back to the parser-generated `PolicyId` (e.g. `"policy0"`).
- Exposes the `@id` label via a new parallel `id_annotations: Dict[str, str]` map on `Diagnostics` and `ValidationResult`, keyed by the parser id. An entry is present whenever the policy declares an `@id` annotation, with the literal value — so `@id("foo")` contributes `"foo"`, and `@id("")` / bare `@id` (which the Cedar docs define as equivalent to `@id("")`) contributes `""`. Policies with no `@id` annotation are omitted from the map.

## Why

Issue #77 reported a regression in the diagnostic surface: in 4.8.1, two policies sharing `@id("baseline-permit-all")` returned distinct `reasons` (`"policy0"` vs `"policy1"`); in 4.8.2 they collapsed to the same string, silently misrouting in callers that used `reasons` as the lookup key into their own `{positional_id → caller_id}` map.

This PR keeps the 4.8.2 ergonomics gain — recover the `@id` label without rebuilding the policy set — while restoring the 4.8.1 disambiguation property.

## Shape of the change

```python
result = is_authorized(request, policies, entities)
result.diagnostics.reasons              # ["policy0"]  — parser-generated id, stable per policy
result.diagnostics.id_annotations       # {"policy0": "baseline-permit-all"}  — present iff @id declared

vr = validate_policies(policies, schema)
vr.errors[0].policy_id                  # "policy0"
vr.id_annotations                       # {"policy0": "baseline-permit-all"}
```

Implementation notes:
- `reason` is now typed `HashSet<PolicyId>` (was `HashSet<String>`) — JSON serialization unchanged (`PolicyId` emits a string), but the Rust type is tighter now that the field no longer holds annotation values.
- `lookup_id_annotation` replaces `resolve_display_id`; it is purely an existence check on the `@id` annotation key with raw `&str` matching (no `PolicySet::annotation` identifier-parse cost on the hot path).

## Behavior of `id_annotations`

| policy declares | `id_annotations[pid]` |
|---|---|
| `@id("foo")` | `"foo"` |
| `@id("")` | `""` |
| `@id` (no parens, per docs ≡ `@id("")`) | `""` |
| no `@id` | omitted |

## Test plan

- [x] Unit tests rewritten for the new shape; new `test_issue_77_multi_tenant_baseline_permit_disambiguates` exercises the reproducer from the issue.
- [x] `pytest tests/unit` — 63 passed, 2 subtests passed.
- [x] `make benchmark-compare` (release-mode build, 5% median / 15% mean gates) — 26 passed, no regression. Largest movement is `test_large_entity_set` at ~4% **faster** than baseline; everything else within ±2% on median.

## Closes

Closes #77.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


skuenzli:
What do you think about this approach @aashitk and @Iamrodos ?

Iamrodos:
See my commend in #77 suggesting **reasons_ids** (growing as my preferred) or reasons_as_ids

skuenzli:
re **reasons_ids**:

from https://github.com/k9securityio/cedar-py/issues/77#issuecomment-4445424876
> Good approach. As policies as passed, id_annotations could speak to a parsing use case or any number of other things. I would prefer to see reasons in the name so it's linked to what the values are. Potentially reasons_as_ids or reasons_ids.

Thank you. I think including 'reason` in the name makes a lot of sense to connect with related data.

I brainstormed names that would connect the id annotations to authz diagnostic reasons and validate policies in the most Pythonic way and settled on:

* `Diagnostics.id_annotations_by_reason`
* `ValidationResult.id_annotations_by_policy_id` (this validation result is always for policy validation)

skuenzli:
Merging to restore unambiguous (policy id) `reasons` and support improved ergonomics with supplemental `Diagnostics.id_annotations_by_reason` and `ValidationResult.id_annotations_by_policy_id` fields.

Issue #80, opened 2026-05-15
Add Cedar policy analysis support


lidaniel7:
Would you be open to adding Cedar policy analysis functions to cedarpy? This would expose cedar-policy-symcc's SMT-based verification capabilities through the existing Python bindings.

Cedar analysis is currently only accessible via Rust or the cedar-lean-cli reference CLI. There are no Python bindings for analysis in any package today. 

To implement this, would add:

- `cedar-policy-symcc = "0.4"` and tokio as Rust dependencies
- Two #[pyfunction]s in `src/lib.rs` following the existing code patterns
- Python wrapper classes (AnalysisResult, CompareResult) in `cedarpy/__init__.py`

I'd be happy to open a PR if interested!

lidaniel7:
And on a separate note, I can't seem to join this slack workspace [Cedar Policy Slack workspace](https://communityinviter.com/apps/cedar-policy/cedar-policy-language) because the invite limit has bene reached

skuenzli:
Hi @lidaniel7 ! Thank you for opening an issue asking whether we should add support for Cedar policy analysis functions to cedarpy.

Ultimately, if a proposed feature is useful to a broad set of users or potential users of cedarpy, then I'm open to adding it as long as it doesn't introduce problems with the core function of the library which is to make Cedar Policy core (engine) available to Python users, particularly  to evaluate whether access is authorized.

I'm generally familiar with Lean and SMT, but I've never used `cedar-policy-symcc`

Can you please briefly describe the specific use cases and developer/user outcomes that you think cedarpy should support with `cedar-policy-symcc`? And for each use case:

* what is your workaround for these use cases today?
* where will the unit+integration tests for those use cases come from; in particular, do test cases already exist in a Cedar Policy-managed repository that we can leverage?

I can see where having symcc available may be able to help cedarpy users improve their own application's policy editing and checking experiences.

From a dependency perspective, can you please point me to any Cedar Policy discussions, RFCs, or PRs that explain why `cedar-policy-symcc` is a distinct package from `cedar-policy`?

The Cedar Policy team decided not to package them together and I'd like to understand why.

I'm definitely wary of picking up additional (transitive) dependencies that would require library consumers to update more frequently because of e.g. a vulnerability in tokio.

lidaniel7:
Hi @skuenzli sorry for the delay! I originally needed this for a project I was working on, but don't need it anymore (or not for a while). Thanks for getting back to me!

[...]

PR #81, opened 2026-05-15
feat: add Cedar policy analysis support

lidaniel7:


skuenzli:
Discussing this in #80

skuenzli:
Closed per #80. We can reconsider this if / when requested.

PR #82, opened 2026-05-21
feat: Add support for partial evaluation

### LINKED ISSUE #28: Support for partial evaluation?
Hi, thank you for your work on this project, it’s been super helpful! Just wondering if there are any plans to support partial evaluation?

swenger:
## Summary

Adds `is_authorized_partial(request, policies, entities, schema=None)`, a Python binding for Cedar's partial-evaluation authorizer. Request fields that are `None` or absent are treated as unknowns; the authorizer returns `Decision.Allow` / `Decision.Deny` when the unknowns can't change the outcome, or `Decision.NoDecision` plus residual policies (as Cedar JSON) for the caller to re-evaluate once unknowns are bound.

`PartialAuthzResult` mirrors `AuthzResult`'s shape (`decision` / `correlation_id` / `diagnostics` / `metrics`) and adds `may_be_determining`, `must_be_determining`, `nontrivial_residuals`, `unknown_entities` to diagnostics. There is no engine version change. cedarpy continues to use cedar-policy 4.8.2 with the `partial-eval` Cargo feature enabled.

The docstring and CHANGELOG both warn that partial-eval results are not a final authorization decision: callers must re-run `is_authorized` once unknowns are bound, since schema type-checking (including action-typed context shapes) is skipped while fields remain unknown.

Fixes #28.

## Test plan

- [x] `make integration-tests` passes
- [x] `pytest tests/unit -v` — full suite passes, including 27 new tests in `tests/unit/test_authorize_partial.py` covering: unknowns at each request slot, definitive Allow/Deny with unknowns, schema interactions, errored-policy diagnostics, residual JSON structure, `@id` annotation surfacing (incl. on errored policies), and three "progressive request filling" cases that exercise partial → complete request transitions
- [x] `make benchmark-compare` — no regression on the post-merge benchmark gate
- [x] Manually verify a residual returned from `is_authorized_partial` can be inspected and acted on by a caller


skuenzli:
@swenger - thank you for this PR. I will take a look when I can. That may be in a couple days.

Can you share how you're using or planning to use this in your own application?

swenger:
Thanks @skuenzli!

> Can you share how you're using or planning to use this in your own application?

Sure! We want to control access to data that's stored in a SQL database. Especially for queries returning more than one result, that means that at least some of the access control needs to happen in SQL (keeping only the rows the user is allowed to see). In order to keep the amount of work in the `WHERE` clause manageable, all parts of the policy set that do not depend on the database row are going to be evaluated before building the query (using the code from this PR), and only the residuals that depend on the row are translated to SQL. We're likely going to perform one last pass with the full policy set on the returned rows 1) as defense in depth against possible mistakes in our SQL implementation and 2) to evaluate any conditions that depend on rows but cannot be translated to SQL.

Currently, implementing this would require starting the cedar CLI in a new process, which isn't great for latency and scalability (besides not being very convenient API to code against).

Does this sound reasonable?

skuenzli:
Thank you @swenger - your use case is great context and I understand it completely -- to the extent I need to ;)

Cargo.toml
```diff
@@ -14,8 +14,8 @@ crate-type = ["cdylib"]
 [dependencies]
 pyo3 = "0.27.0"
 anyhow = "1.0"
-cedar-policy = "4.8.2"
-cedar-policy-cli = "4.8.2"
-cedar-policy-formatter = "4.8.2"
+cedar-policy = { version = "4.10.0", features = ["partial-eval"] }
+cedar-policy-cli = "4.10.0"
+cedar-policy-formatter = "4.10.0"
```
skuenzli: Upgrades to cedar-policy engine need to be done separately from the addition of features like partial evaluation.

What prompted you to upgrade to 4.10.0?

Interestingly, the tests pass with cedar-policy 4.8.2:
```
cedar-policy = { version = "=4.8.2", features = ["partial-eval"] }
cedar-policy-cli = "=4.8.2"
cedar-policy-formatter = "=4.8.2"
```

cedarpy/__init__.py
```diff
@@ -253,6 +253,121 @@ def policies_from_json_str(policies: str) -> str:
...
+        return self._resp.get('residuals', {})
+
+    @property
+    def id_annotations(self) -> dict:
+        return self._resp.get('id_annotations', {})
+
+    @property
+    def diagnostics_errors(self) -> List[str]:
+        return list(self._resp.get('diagnostics_errors', []))
+
```
skuenzli, line 336: `PartialAuthzResult` is shaped differently from `AuthzResult`. In particular:

* the `Diagnostics` are flattened onto `PartialAuthzResult`
* id_annotations instead of `id_annotations_by_reason` which is now on `Diagnostics` (note: the name changed from `id_annotations` proposed in the original PR)

Can you please help me understand the reasoning there? 

It looks like you've mirrored a lot of cedar-policy's `PartialResponse` shape, which is a sensible place to start. Especially since mirroring `PartialResponse` should align cedar-py partial eval to the engine's own semantics and documentation.

However, I think the result is confusing in places, especially for `diagnostics_errors` and `errored`.

What do you think about adopting a structure closer to `cedar-java`'s [`PartialAuthorizationResponse`](https://github.com/cedar-policy/cedar-java/blob/main/CedarJava/src/main/java/com/cedarpolicy/model/PartialAuthorizationResponse.java#L40)

What are your thoughts as someone who actually built on top of the Python partial evaluation API?

cedarpy/__init__.py
```diff
@@ -253,6 +253,121 @@ def policies_from_json_str(policies: str) -> str:
...
+                          schema: Union[str, dict, None] = None,
+                          verbose: bool = False) -> PartialAuthzResult:
+    """Partially evaluate an authorization request with unknowns.
+
+    Fields in the request dict that are None or absent are treated as
+    unknown. The evaluator simplifies policies as far as possible and
+    returns residual expressions for policies that cannot be fully resolved.
+
+    :param request: dict with optional keys: principal, action, resource,
+        context, correlation_id. Any key that is None or absent is unknown.
```
skuenzli: I think we want to add a subtle clarification after 'Any key that is None or absent is unknown':
> Unlike is_authorized (which defaults an absent context to empty), an absent or None context here is treated as unknown and will residualize; pass context={} for an explicitly empty context.

tests/unit/test_authorize_partial.py
```diff
@@ -0,0 +1,281 @@
...
+        request={
+            "principal": 'User::"alice"',
+            "action": 'Action::"view"',
+            "resource": 'Photo::"photo1"',
+            "context": {"level": 10},
+        },
+        policies=policies,
+        entities="[]",
+    )
+    assert result.decision == Decision.Allow
```
skuenzli, line 451: Are there any test cases that we can adopt from the cedar-policy project verifying

It looks like there are gaps for:

1. schema= tests: Partial eval interacts with schema (action-typed contexts, request validation). Add at least one passing and one failing schema test.

 2. definitely_errored test. No policy in any test errors at evaluation time. Easy   to construct: permit(...) when { resource.missing_attr };.

 3. must_be_determining / may_be_determining / errored are never asserted to be  non-empty. All assertions are presence-only or empty.

4. test_residuals_are_valid_cedar_syntax is pretty loose. the test is fully deterministic, so maybe just assert the result it produces. I definitely wouldn't expect to see `forbid` in the residual.

skuenzli (changes requested):
Thank you for proposing an implementation to support partial evaluation. Overall, I think is_authorized_partial uses a good approach.

I've added a several comments and requests.

If we can work through those, then I think we can land this change.

Cargo.toml
swenger: Weird, I remember it wasn't working with 4.8.2 and started working once I bumped the version... But you're right, the tests absolutely work with 4.10.0 and there's nothing in the changelog that suggests otherwise. I've rolled the upgrade back in https://github.com/k9securityio/cedar-py/pull/82/changes/ef349d59abffb636212cfd04198dadb331da2388.

I presume adding `features = ["partial-eval"]` here is fine though?

As far as I can tell, we _will_ [need to upgrade to 4.11.0](https://github.com/cedar-policy/cedar/blob/92b97a45c091bb285b997664bf14bfffd8c402cc/cedar-policy/CHANGELOG.md?plain=1#L22) if we want to support partial evaluation of templates, but given that we [don't have support for regular evaluation of templates yet](https://github.com/k9securityio/cedar-py/issues/29), I don't think it makes sense to bump the version now.

cedarpy/__init__.py
swenger: Good point, thanks. If added your suggestion in https://github.com/k9securityio/cedar-py/pull/82/changes/6a5e832049fd4bdc390c99165052695106660895 and also aligned the rest of the docstring with the one of `is_authorized` so the differences are visible at a glance.

tests/unit/test_authorize_partial.py
swenger: Good catch - I've added tests in https://github.com/k9securityio/cedar-py/pull/82/changes/6a3324e4ffaf1bd036727af4b0b6e42226c80338:

1. Partial eval with schema:
    1. `test_partial_with_schema_unknown_principal` asserts that a missing `principal` passes schema validation but ends up in the residual.
    2. `test_partial_with_schema_all_known` asserts that a correct `principal` passes schema validation and the policy is satisfied.
    3. `test_partial_with_schema_wrong_principal_type` asserts that an incorrect `principal` results in an error and the `decision` is `None`.
    4. `test_partial_with_schema_unknown_context` checks the special case of missing `context`, asserting that the context passes schema validation but the required context field ends up in the residuals.
2. `test_definitely_errored_type_mismatch` asserts that `permit(...) when { resource.missing_attr }` puts the affected policy in `errored`, but other policies still evaluate correctly.
3. `must_be_determining` / `may_be_determining`:
    1. `test_may_be_determining_specific_ids` asserts that two `permit` policies that could still become satisfied end up in `may_be_determining`.
    2. `test_must_be_determining_with_definitive_decision` asserts that when a `forbid` policy is satisfied, it ends up in `may_be_determining` and `must_be_determining`.
    3. `test_determining_sets_exclude_irrelevant` asserts that a satisfied policy ends up in `may_be_determining` and `must_be_determining`, and an unsatisfied policy ends up in neither.
4. Replaced `test_residuals_are_valid_cedar_syntax` with the following:
   1. `test_residual_for_principal_check` asserts residual condition for unknown `principal` (no `when` in original policy).
   2. `test_residual_for_context_condition` asserts residual condition for unknown `context` (used in `when` in original policy).

Is this more or less what you meant?

cedarpy/__init__.py
swenger: Yes, I agree it's confusing, especially for someone already familiar with `cedarpy.AuthzResult`. Consistency with the rest of the Python API is probably more important than consistency with the Rust API. I'll see if I can refactor this to make more sense - I'll get back to you when I'm done.

How would you handle "user error" like policies that cannot be parsed or entities that do not correspond do the schema? Raise `ValueError` like [some of the existing code seems to do](https://github.com/k9securityio/cedar-py/blob/d4844ad2754b8f603f03325d770f9312cf5a7031/cedarpy/__init__.py#L251)? I'd rather not hide this in the diagnostics.

swenger: I've [refactored](https://github.com/k9securityio/cedar-py/pull/82/changes/6a3324e4ffaf1bd036727af4b0b6e42226c80338..c846135d936375eac4fdad779c4251853a39c66e) the `is_authorized_partial` interface for maximum consistency with the existing `is_authorized`, including how errors are handled - the user will need to check `decision` _and_ (if `decision` is not `Allow` or `Deny`) `residuals` to determine whether the request was successful, and this behavior is documented in `PartialAuthzResult`'s docstring. As much as I would prefer raising on error, I think being consistent across the API is more important.

The `residuals` are returned in Cedar's native JSON format (as a Python dict), which can be converted back to Cedar policy language using `json.dumps` and `policies_from_json_str`. For our use case, the parsed format is most convenient.

Does this look reasonable to you?

Cargo.toml
skuenzli: Super - thank you. Enabling the partial-eval feature is fine.

Thanks for the note on requiring 4.11 to evaluate partials in templates.

Cargo.toml
```diff
@@ -14,7 +14,7 @@ crate-type = ["cdylib"]
 [dependencies]
 pyo3 = "0.27.0"
 anyhow = "1.0"
-cedar-policy = "4.8.2"
+cedar-policy = { version = "4.8.2", features = ["partial-eval"] }
```
skuenzli, line 17: 👍

tests/unit/test_authorize_partial.py
skuenzli: Thank you. This is what I meant.

cedarpy/__init__.py
skuenzli: The refactored `is_authorized_partial` interface and `PartialAuthzResult` aligns to the existing `AuthzResult` and how I was thinking about it originally. 

But we need to find some place for the additional data returned by the partial call such as `satisfied` and `must_be_determining`. I see three locations, though I'm sure there are more:
1. directly on `PartialAuthzResult`
2. on a `PartialDiagnostics` that supplements `Diagnostics` with the new fields
3. on a `PartialEvalDetails` object that is populated when evaluation of the partial succeeds.

swenger: By `satisfied`, do you mean [`definitely_satisfied`](https://github.com/cedar-policy/cedar/blob/92b97a45c091bb285b997664bf14bfffd8c402cc/cedar-policy/src/api.rs#L1174)? That's already exposed via `Diagnostics.reason`.

I added `may_be_determining`, `must_be_determining`, `nontrivial_residuals`, and `unknown_entities` as a `PartialDiagnostics` subclass of `Diagnostics` in https://github.com/k9securityio/cedar-py/pull/82/changes/20fb68d12f658316010d2e85e7fd6b9463cb0718. Maybe I'll also change the current `residuals` to only contain the `nontrivial_residuals` and store `all_residuals` in `PartialDiagnostics` instead - I believe most use cases probably only care about the nontrivial residuals. The reasoning above about error handling should still hold: If there are no `nontrivial_residuals`, the `decision` is known, so empty `nontrivial_residuals` plus unknown `decision` means there was an error.

swenger: Actually, if we want to make the `nontrivial_residuals` the default, we'd duplicate the residual information; if we keep `all_residuals`, then `nontrivial_residuals` can simply contain the residual IDs and their values can be looked up in `all_residuals`. So maybe let's leave it this way.

src/lib.rs
```diff
@@ -559,12 +570,197 @@ fn validate_policies(policies: String, schema: String) -> String {
...
+        let error_map: HashMap<&PolicyId, String> = concretized.diagnostics().errors()
+            .map(|e| {
+                let AuthorizationError::PolicyEvaluationError(pe) = e;
+                (pe.policy_id(), e.to_string())
+            })
+            .collect();
+        errored_ids.iter()
+            .map(|pid| error_map.get(pid).cloned()
+                .unwrap_or_else(|| format!("error while evaluating policy `{pid}`: unknown error")))
+            .collect()
```
swenger, line 712: @skuenzli I had to add this to extract the [error messages](https://github.com/cedar-policy/cedar/blob/92b97a45c091bb285b997664bf14bfffd8c402cc/cedar-policy-core/src/authorizer/partial_response.rs#L411) since the [public API](https://github.com/cedar-policy/cedar/blob/92b97a45c091bb285b997664bf14bfffd8c402cc/cedar-policy/src/api.rs#L1179) unfortunately only returns policy IDs.

If we assume most authz requests don't cause errors, this comes at no additional runtime cost. If it's likely that "regular" requests will contain errors in some policies, we'll have to make a tradeoff - 1) concretize and evaluate again to get the error messages (implemented here), 2) live with having only the erroring policy IDs (and potentially manually make another call to `is_authorized` to get error messages), or 3) try to extend the public API upstream to expose the error messages directly.

Let me know what you prefer and I'll document our decision and reasoning in a comment in the code.

skuenzli:
@swenger - thanks for the updates and your patience. I wasn't able to review the changes today, but will try to review it tomorrow.

cedarpy/__init__.py
skuenzli: Yes, I was referring to `definitely_satisfied`, `definitely_errored`, and the several that you've added to `PartialDiagnostics`

I like the members of `PartialDiagnostics` exposing residuals.

If you don't think we need dedicated `definitely_satisfied` and `definitely_errored` fields because users can consult the `reasons` field, then I think we can start with that. We can always add more properties.

The only thing I'm still thinking about here is whether PartialDiagnostics should extend Diagnostics or if they should be independent siblings with similar 'base' properties. `Diagnostics` didn't change in almost 3 years except for the recent `id_annotations_by_reason` property, which applies to both. So maybe it's the right relationship: PartialDiagnostics specializes Diagnostics.

src/lib.rs
skuenzli: I think the approach is good. I added a comment to try to help explain what is happening. Feel free to improve it.

skuenzli:
No changes in performance of existing codepaths were expected.

The existing benchmark suite passes, verifying that:

```
===== HEAD: cae8ad3 Merge branch 'main' into swenger-partial =====
===== RUNS=5  OUT_DIR=tests/benchmark/results/current =====
---- building (release) at 22:47:43 ----
📦 Built wheel for CPython 3.11 to /var/folders/k2/tnw8n1c54tv8nt4557pfx3440000gp/T/.tmpGYw3js/cedarpy-4.8.3-cp311-cp311-macosx_11_0_arm64.whl
✏️ Setting installed package as editable
🛠 Installed cedarpy-4.8.3
  [22:47:45] run 1 -> tests/benchmark/results/current/run1.json
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================= 26 passed in 24.74s ==============================
  [22:48:10] run 2 -> tests/benchmark/results/current/run2.json
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================= 26 passed in 25.19s ==============================
  [22:48:36] run 3 -> tests/benchmark/results/current/run3.json
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================= 26 passed in 25.07s ==============================
  [22:49:01] run 4 -> tests/benchmark/results/current/run4.json
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================= 26 passed in 24.46s ==============================
  [22:49:26] run 5 -> tests/benchmark/results/current/run5.json
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
============================= 26 passed in 24.54s ==============================
===== DONE at 22:49:51 =====

Benchmark                            Baseline (μs)  Median (μs)   Δ median  Status
-----------------------------------  -------------  -----------  ---------  ------
test_batch_complex_policy                     1234         1283      +4.0%  PASS
test_batch_simple_policy                       384          385      +0.2%  PASS
test_batch_size_scaling[100]                  4112         4034      -1.9%  PASS
test_batch_size_scaling[10]                   1105         1072      -3.0%  PASS
test_batch_size_scaling[1]                     799          789      -1.3%  PASS
test_batch_size_scaling[25]                   1595         1572      -1.4%  PASS
test_batch_size_scaling[50]                   2430         2374      -2.3%  PASS
test_batch_size_scaling[5]                     933          912      -2.3%  PASS
test_complex_policy                            273          269      -1.5%  PASS
test_context_as_dict                           802          790      -1.6%  PASS
test_context_as_json_string                    799          790      -1.1%  PASS
test_entities_as_json_string                   727          710      -2.4%  PASS
test_entities_as_list                          796          788      -1.0%  PASS
test_large_entity_set                         3875         3930      +1.4%  PASS
test_medium_entity_set                         795          795      +0.0%  PASS
test_medium_policy                             187          184      -1.3%  PASS
test_sandbox_b_batch_mixed_actions            1098         1115      +1.5%  PASS
test_sandbox_b_batch_multiple_users            683          684      +0.1%  PASS
test_sandbox_b_delete_with_auth                574          573      -0.1%  PASS
test_sandbox_b_view_own_photo                  574          585      +1.9%  PASS
test_sandbox_b_view_public_photo               576          588      +2.1%  PASS
test_simple_policy_allow                       144          142      -1.5%  PASS
test_simple_policy_deny                        143          143      -0.3%  PASS
test_small_entity_set                          187          188      +0.4%  PASS
test_ten_batch_call                           1098         1079      -1.7%  PASS
test_ten_single_calls                         8243         8121      -1.5%  PASS

N=5 current runs vs tests/benchmark/results/baseline.json; threshold: median Δ > 5% fails
RESULT: PASS
```

skuenzli (commented):
Thanks for making the requested adjustments! I have also clarified and fixed a few things. I think this PR is ready to merge.

Notably, I restored the Cargo.lock from main to avoid picking up some dependency changes that were introduced early in your branch. The result is that there are no dependency changes except enabling cedar-policy 4.8.2's `partial-eval` feature.

Please review the full PR including my changes. Then either resolve or add/update conversations if you think we need to make more changes.

cedarpy/__init__.py
swenger: Great, thanks!

> `Diagnostics` didn't change in almost 3 years except for the recent `id_annotations_by_reason` property, which applies to both. So maybe it's the right relationship: PartialDiagnostics specializes Diagnostics.

I think so too, and it allows callers to potentially reuse code for both partial and final authorization when they don't need the additional fields (not that it makes a huge difference in Python).

src/lib.rs
swenger: Thanks! I think the comment is perfect.

cedarpy/__init__.py
```diff
@@ -328,6 +332,23 @@ def is_authorized_partial(request: dict,
...
+        ``is_authorized`` with the complete request; that call performs
+        full schema validation (including action-typed context shapes)
+        which partial evaluation deliberately skips for fields that are
+        still unknown.
+
+        In particular, when a schema is provided but ``action`` is
+        unknown, request-context type-checking against the schema's
+        action-specific context shape is silently skipped — there is no
+        bound action to look up. A residual that depends on context
+        values is not a guarantee that those values are well-typed.
```
swenger, line 371: Just to be sure: _If_ everything is well-typed (which could be enforced Python-side using pydantic models, type annotations, etc.), a partial-eval decision (if there is any) should still be reliable, right?

src/lib.rs
```diff
@@ -698,9 +698,17 @@ fn is_authorized_partial(
         reason.insert(pid);
     }
     let errored_ids: Vec<&PolicyId> = partial_response.definitely_errored().collect();
+    for pid in &errored_ids {
+        if let Some(annotation) = lookup_id_annotation(&policy_set, pid) {
+            id_annotations_by_reason.insert(pid.to_string(), annotation);
+        }
+    }
```
swenger: Shouldn't _all_ policy IDs be returned by `all_residuals()` below? If so, we might be able to remove this block as well as the corresponding one for `definitely_satisfied()` above.

I've implemented this in https://github.com/k9securityio/cedar-py/pull/82/changes/63e84d469b65ec9165ca4e6ba4613bef27bc7d80 - let me know if you disagree.

swenger:
Thanks @skuenzli! I've made one minor change for you to review (see https://github.com/k9securityio/cedar-py/pull/82#discussion_r3316659977) and left another question (https://github.com/k9securityio/cedar-py/pull/82#discussion_r3316632552), but generally all looks good to me. Feel free to merge if you agree!

I've updated my SQL generation code to use the latest version (with nontrivial residuals etc.) and it seems to work fine 👍

cedarpy/__init__.py
skuenzli: I think the caveat provides the correct guidance.

I just added a [`test_partial_with_schema_unknown_action_skips_context_type_check`](https://github.com/k9securityio/cedar-py/pull/82/changes#diff-1c9fabe87f344c3b59a1780d4fc5f74dbe5d0ee220cc529a809ae26fca47bbd4R613) that demonstrates the problem concretely.

LMK what you think. It's not that I think the partial-eval decision is unreliable (I trust cedar and its model). More that users might be surprised by a an `Decision.Allow` result decision when they don't specify an action.

src/lib.rs
skuenzli: yes, I think your change is correct. thank you.

skuenzli:
> Thanks @skuenzli! I've made one minor change for you to review (see [#82 (comment)](https://github.com/k9securityio/cedar-py/pull/82#discussion_r3316659977)) and left another question ([#82 (comment)](https://github.com/k9securityio/cedar-py/pull/82#discussion_r3316632552)), but generally all looks good to me. Feel free to merge if you agree!
> 
> I've updated my SQL generation code to use the latest version (with nontrivial residuals etc.) and it seems to work fine 👍

That's terrific! Thank you for testing.

cedarpy/__init__.py
skuenzli: I'm still feeling reluctant to binding PartialDiagnostics to Diagnostics, mostly for the long-term constraints it places on `Diagnostics`.

What do you think about using a shared private base class?

1. extract `Diagnostics`' current members to `_DiagnosticsBase`
2. `Diagnostics` extends `_DiagnosticsBase`; doesn't add anything
3. `PartialDiagnostics` extends `_DiagnosticsBase`; adds specializations

```python
class _DiagnosticsBase:
    """Shared backing for the public diagnostics types. Not part of the
    public API — type-annotate against ``Diagnostics`` or
    ``PartialDiagnostics`` directly. Exists to share dict-accessor
    implementation without committing the two public types to a
    subclass relationship.
    """

    def __init__(self, diagnostics: dict) -> None:
        super().__init__()
        self._diagnostics: dict = diagnostics

    @property
    def errors(self) -> List[str]:
        return self._diagnostics.get('errors', list())

    @property
    def reasons(self) -> List[str]:
        # (intentionally) map 'reason' key in diagnostics dict to 'reasons' property (plural)
        return self._diagnostics.get('reason', list())

    @property
    def id_annotations_by_reason(self) -> dict:
        """Map from each parser-generated policy id in ``reasons`` to the
        literal value of its ``@id`` annotation, when the matched policy
        declares one. ``@id("foo")`` contributes ``"foo"``; ``@id("")`` /
        bare ``@id`` (which the Cedar docs define as equivalent to
        ``@id("")``) contributes ``""``. Policies with no ``@id`` annotation
        are omitted from the map.
        """
        return self._diagnostics.get('id_annotations_by_reason', dict())

class Diagnostics(_DiagnosticsBase):
    """Diagnostics for a fully-evaluated authorization decision."""
    pass


class PartialDiagnostics(_DiagnosticsBase):
    """Diagnostics for a partial-evaluation authorization decision.

    Carries the same ``errors`` / ``reasons`` / ``id_annotations_by_reason``
    surface as ``Diagnostics``, plus partial-eval-specific fields. Note that
    semantics differ slightly: in partial eval, ``reasons`` lists the
    definitely-satisfied policies, and ``id_annotations_by_reason`` covers
    annotations for definitely-satisfied, all-residual, and
    definitely-errored policies.
    """

    @property
    def may_be_determining(self) -> List[str]:
        return self._diagnostics.get('may_be_determining', [])

    @property
    def must_be_determining(self) -> List[str]:
        return self._diagnostics.get('must_be_determining', [])

    @property
    def nontrivial_residuals(self) -> List[str]:
        return self._diagnostics.get('nontrivial_residuals', [])

    @property
    def unknown_entities(self) -> List[str]:
        return self._diagnostics.get('unknown_entities', [])

```

All existing tests pass with this implementation.

cedarpy/__init__.py
swenger: Ah I see, I missed the "action-specific" part. So the problem arises when neither the action nor the context value influence the outcome, but the context value would cause a schema violation error _if the relevant action-specific schema were known_.

I would still hope that this doesn't show up in "regular operation" since the application would presumably construct the correct type of context once the action is known, and therefore wouldn't even be able to construct _any_ action-dependent context yet since in the failure scenario the action is _not_ known. But definitely a surprising edge case users should be aware of.

I've tightened the assertions in the test a little to show that no errors _of any kind_ are detected, not just no "type" errors - when action is provided, the error message doesn't contain "type" either (unless schema validation is deactivated and the non-boolean value is used in a boolean context). I've also added the corresponding "action known" test case for comparison: https://github.com/k9securityio/cedar-py/pull/82/changes/4ecdc62d30739e4e3d712d7f980b4e2fc7d6014d

cedarpy/__init__.py
swenger: I'd be OK with this, but it's a lot of not immediately obvious logic mostly just to avoid copying some code, right? Since callers that want do handle either partial or final evaluation would not use the (private) `_DiagnosticsBase` (as a type annotation), I don't think there's much to be gained from having it in the first place. Simply copying the entire `Diagnostics` class body into `PartialDiagnostics` (and not having any inheritance) would take up more space, but be straightforward to understand and less likely to cause accidental breakage if any edits are ever made to the base class that should actually be in one of the derived classes.

But I don't think that's a huge problem, let me implement your solution for now and then you tell me whether to keep it or copy the definition :)

swenger: https://github.com/k9securityio/cedar-py/pull/82/changes/c309fba0e2a3c924f031242bafad6b9d5c916594

skuenzli: yes... I considered duplicating the impl but was reluctant due to drift.

the other option I considered is composing a Diagnostics impl inside PartialDiagnostics, then delegating the shared method calls.

Thanks for integrating my suggestion.

Let's move forward.

cedarpy/__init__.py
skuenzli: Yes, that's exactly it.

Thanks for adding `test_partial_with_schema_unknown_action_skips_context_type_check_when_complete`

skuenzli (approved):
This addition looks great. Thank you for your contribution and patience iterating on the solution.

swenger:
Thanks @skuenzli for the review and great collaboration! I think what we merged is indeed much better than what I started out with ❤️

skuenzli:
@swenger - I'm so glad you feel that way. I think so too!

"Simple" isn't easy!

PR #87, opened 2026-05-29
Run the upstream fuzzer corpus suite (59,696 cases) and add the {type, id} EUID form needed to make them pass

geekphilosophy:
Summary

  This PR does three related things, in three commits:

  1. test: compare reason/errors as sets, unskip all integration tests - the existing make integration-tests runner compared the reasons and errors lists as ordered sequences, which caused 5 of the 74 hand-written integration cases to be @unittest.skip'd. Cedar's authorization decision is contractually a set of contributing policies, not an ordered list — the upstream Rust runner and cedar-java both compare as sets. Switching to set() comparison and removing the skips brings cedar-py to 74/74 on the hand-written suite.
  2. feat: accept structured {type, id} EUIDs in authorization requests - is_authorized and is_authorized_batch now accept either a Cedar surface-syntax string ('User::"alice"', unchanged) or a structured dict ({'type': 'User', 'id': 'alice'}) for principal / action / resource. The dict form routes through EntityUid::from_json and tolerates entity ids containing whitespace, control chars, or other characters Cedar's surface parser rejects with "needs to be normalized". This mirrors cedar-java's JsonEUID form, and is a pure addition - string-form callers see no change.
  3. test: add corpus-tests target running the upstream fuzzer corpus - a new make corpus-tests target runs the 7,462-file fuzzer-generated corpus shipped alongside cedar-integration-tests (59,696 individual request cases). Kept as a separate target - not part of make integration-tests - because it takes ~5 minutes. The runner adopts the leniency rules used by the upstream Rust runner and cedar-java's SharedIntegrationTests: reasons compared as a set, errors compared by count, and NoDecision (parse failure) soft-passes when the fixture expects Deny. The dict-form EUID support from commit 2 is what lets these pass - fuzzer-generated entity ids are routinely not expressible in Cedar surface syntax.

  Motivation

  The README claimed cedar-py passed "69 of 74" hand-written integration tests; investigating turned up that all 5 "failures" were actually correct decisions failing on incidental ordering differences. While fixing that, it became clear cedar-py wasn't running the fuzzer corpus at all - the upstream Cedar repo ships ~60K request cases that cedar-policy (Rust) and cedar-java run as part of CI, and cedar-py was leaving them on the table.

A first cut at the corpus runner produced ~92% pass / 4,746 fail, all sharing one root cause: the test fixture's entity ids contain characters Cedar's surface-syntax parser rejects, and cedar-py's only on-ramp was a Cedar-syntax string. cedar-java sidesteps this with a structured JsonEUID form. Commit 2 brings the same path to cedar-py - which is the canonical way Cedar APIs accept entity references when the id may contain arbitrary content (EntityUid::from_json in the cedar-policy crate).

  Testing

  - make integration-tests - 74/74 pass.
  - make corpus-tests - 59,696/59,696 pass in 4:46 wall-clock locally.
  - New unit tests in tests/unit/test_authorize.py:
    - test_principal_action_resource_may_be_string_or_dict - string, dict, and mixed forms in the same request.
    - test_dict_form_euid_accepts_ids_unrepresentable_in_surface_syntax - uses an id with an embedded newline ('J\nK') which the surface parser rejects, confirms the dict form handles it.
  - make quick (existing unit suite + benchmarks) continues to pass.

  Notes

  - README updated in two places: hand-written-suite count corrected to 74, and a new "Corpus tests" subsection documents make corpus-tests.
  - Docstrings on is_authorized and is_authorized_batch document both EUID forms.
  - No backwards-incompatible changes - every existing call site using string-form EUIDs continues to work unchanged.e


skuenzli:
Chris - this sounds fantastic. Thank you!

FWIW, the fuzzer tests didn't exist (or weren't publicly available) when the initial test suites were built.

I will review the change as soon as I can but it may be a couple of days.

I hope to merge this before the next release.

src/lib.rs
```diff
@@ -56,13 +57,43 @@ fn policies_from_json_str(s: String) -> PyResult<String> {
...
+        match self {
+            EntityUidInput::Cedar(s) => s
+                .parse()
+                .context(format!("Failed to parse {what} as entity Uid")),
+            EntityUidInput::Json(v) => EntityUid::from_json(v.clone())
+                .context(format!("Failed to parse {what} as entity Uid (JSON form)")),
+        }
+    }
+}
+
```
skuenzli, line 88: this is neat.

skuenzli:
Thank you for this PR! It's awesome. 

**Adjustments to tests**

I made a couple minor adjustments to the tests: 

* `49e88c8e` - move corpus extraction out of the shared conftest: The extraction and param-building ran at module level in `tests/integration/conftest.py`. A conftest loads for everything in its directory, so `make integration-tests` (which runs only the hand-written suite) was unpacking the 4.1 MB / 7,462-file tarball and parsing every JSON just to collect tests — about 10.4s of wall time. It now lives in `test_cedar_corpus_tests.py`, so that work only runs when the corpus module is collected. Hand-written collection drops to ~0.55s, and `make corpus-tests` is unchanged.

* `ebe2933e` - fix the `_extract_corpus_once` return type: It returns `None` when the tarball is missing but was annotated `-> Path`. Changed to `-> Optional[Path]`.

**Backwards compatibility**

I specifically reviewed the changes for backwards-compatibility. Claude and I found:

The Python API doesn't change; only docstrings do. The native crate is `cdylib`-only and isn't published to crates.io, so the Rust signature changes (`EntityUidInput`, `RequestArgs`) have no external consumer. String-form EUIDs follow the same `from_str` path as before, and a bad EUID still produces a per-request `NoDecision`. The `{type, id}` dict form is a new option, not a replacement. One behavior does change: a request missing `principal`, `action`, or `resource` now raises a clean `KeyError` instead of a Rust panic, which seems strictly better.

Verified locally against a fresh `maturin develop --release` build: 74/74 hand-written integration tests, 59,696/59,696 corpus cases, and the full unit suite (including the two new EUID tests) pass.

skuenzli:
Also fixed up CI to not run the corpus in integration tests. Will file another issue to harmonize the unit/integration test execution across platforms.

skuenzli:
Thank you @geekphilosophy !

PR #91, opened 2026-06-14
feat: reusable PolicySet handle to avoid re-parsing policies on every call 

### LINKED ISSUE #83: perf: expose a reusable PolicySet to avoid re-parsing on every is_authorized call
I’ve been profiling per-request cost in a Lambda-based service that uses cedarpy with a static, code-checked policy set, and the per-call `PolicySet::from_str` inside `is_authorized` is showing up as the dominant cost. On the same machine I’m seeing ~1 ms per call against a 16 KB policy file vs ~0.07 ms with a single-permit policy, so the parse really is most of the work. 

Since `cedar_policy::Authorizer::is_authorized` already takes `&PolicySet` directly, I’m wondering whether you’d be open to exposing a Python-side PolicySet (e.g. `PolicySet.from_str(...)`) that is_authorized and is_authorized_batch could accept alongside str. So backwards compatible. Callers whose policies are static could then parse once and reuse the object for the lifetime of the process. 

I would be happy to create a PR if there is interest.

Iamrodos:
## Summary

Adds an opaque, reusable `PolicySet` handle so callers with static policies can parse once and reuse the parsed object across authorization calls, instead of re-parsing the policy text on every `is_authorized*` call. Parsing the policy set is the dominant per-call cost; for long-running services and Lambdas with code-checked, static policies that parse is pure waste on every call.

Closes #83.

## Motivation

Profiling a Lambda-based service with a static, code-checked policy set showed `PolicySet::from_str` inside `is_authorized` as the dominant per-call cost (issue #83). `cedar_policy::Authorizer::is_authorized` already takes `&PolicySet`, so we expose a Python-side handle that holds the parsed set and is borrowed zero-copy across the FFI — the pydantic-core `SchemaValidator` pattern. Design follows the discussion in #83 and is informed by the cedar-java caching impl (cedar-java#350); the API mirrors cedar-java / cedar-policy-rb.

## Usage

```python
from cedarpy import PolicySet, is_authorized, is_authorized_batch

policy_set = PolicySet.from_str(policies)        # parse once (e.g. at cold start)

is_authorized(request, policy_set, entities)     # reuse — no re-parse
is_authorized_batch(requests, policy_set, entities)
```

A handle is accepted anywhere a policies string is accepted: `is_authorized`, `is_authorized_batch`, `is_authorized_partial`. Also `PolicySet.from_json_str(...)` for the Cedar JSON/EST format; supports `len(ps)` and `str(ps)`.

## Design decisions

- **Widen the union, don't add functions.** The `policies` parameter goes from `str` to `str | PolicySet`, keeping the existing call shape — the approach preferred in #83.
- **Opaque, `frozen` `#[pyclass]`** holding the Rust `PolicySet`, borrowed zero-copy on the authz path; memory reclaimed by Rust `Drop` when the last Python ref drops (no manual release), per #83.
- **`from_str` / `from_json_str` factories** — mirrors cedar-java / `cedar-policy-rb` and the existing `policies_*_json_str` helpers; no bare constructor, so the source format is always explicit.
- **Construction raises `ValueError` on bad policies** (eagerly), rather than folding errors into a result. This matches the existing parse helpers (`format_policies`, `policies_to_json_str` / `_from_json_str` already raise `ValueError`) since it's a *constructor*, not an authorization call.
- **Scope: policies only.** Structured so the same pattern (opaque `frozen` pyclass + `str | handle` extractor + borrow-into-slot) can extend to reusable `Entities` / `Schema` handles.

## Behavior changes

| Change | Before | After |
|--------|--------|-------|
| **stdout on errors** | `is_authorized*` printed policy-parse / per-request / serialization errors to stdout (batch: unconditional; partial: `verbose`-only) | printed only when `verbose=True`, matching the rest of the authz path; errors still returned in `diagnostics` |
| **metrics** | — | successful results gain `policies_pre_parsed` (`1` handle / `0` string); `parse_policies_duration_micros` reads ~0 when reusing a handle |

Both are additive / opt-in for string callers; flagged here for visibility. The new metric is a `0/1` value in the existing `metrics` (`HashMap<String, u128>`) map — flag if you'd prefer it as a top-level field.

## Performance

Release build, median `is_authorized` with a ~10-entity set, via `pytest tests/benchmark --benchmark-only -k PolicySetReuse`:

| Policy | string (re-parse) | handle (reuse) | speedup |
|--------|-------------------|----------------|---------|
| simple (1 rule) | ~152 µs | ~120 µs | ~1.3× |
| medium (4 rules) | ~183 µs | ~125 µs | ~1.5× |
| typical (60 rules, ~16 KB) | ~1.54 ms | ~168 µs | **~9.2×** |

The win scales with policy size: on the typical production-scale set, reuse removes ~1.4 ms of policy parsing per call. (Calls still parse entities each time, so it's the *policy* parse that's amortized.) The `typical` row is a synthetic ~16 KB fixture; the same measurement on a real 16 KB / 63-policy production policy gave ~1 ms of parse and a comparable speedup — matching what we profiled when opening #83.

## Test plan

- [x] `maturin develop --release` — clean build
- [x] `pytest tests/unit` — **126 passed** (incl. 23 new in `tests/unit/test_policyset.py`: string-path equivalence, reuse, batch / partial parity, schema, templates, empty set, two independent handles, multithreaded shared-handle, eager `ValueError`, wrong-type `TypeError`, the new metric)
- [x] `make benchmark-compare` — this change is **performance-neutral**: a direct HEAD-vs-`main` A/B on the affected benchmarks is within ±0.6%. (The gate's `baseline.json` was captured at the 4.8.0 release and predates ~20 merged commits — e.g. #87's structured-EUID handling, #82 partial-eval — that added per-call cost, so it reports accumulated drift on `main`, not this change; refreshing the baseline is out of scope here. New reuse benchmarks have no baseline and are skipped.)
- [x] `cedarpy/_internal.pyi` + `py.typed` confirmed shipped in the built wheel

## Follow-ups

Reusable `Entities` / `Schema` handles (same pattern), if there's interest.

---


skuenzli:
@Iamrodos - thank you for the PR. I (also) just returned from traveling for a couple weeks. I will try to review this week.

src/lib.rs
```diff
@@ -56,6 +56,113 @@ fn policies_from_json_str(s: String) -> PyResult<String> {
...
+                        errs.push(Error::msg(err_message));
+                        PolicySet::new()
+                    }
+                };
+                slot.insert(pset)
+            }
+        }
+    }
+}
+
```
skuenzli, line 165: 👍

src/lib.rs
```diff
@@ -143,26 +250,24 @@ fn is_authorized_batch(requests: Vec<Bound<'_, PyDict>>,
...
-        Err(parse_errors) => {
-            let err_message = format!("policy parse errors:\n{:#}",
-                                      parse_errors.to_string());
-            println!("{:#}", err_message);
-            errs.push(Error::msg(err_message));
-            PolicySet::new()
-        }
-    };
+    let mut policy_set_slot: Option<PolicySet> = None;
+    let policy_set = policies.resolve(&mut policy_set_slot, &mut errs, verbose);
```
skuenzli, line 270: 👍

tests/benchmark/test_benchmark_authorize.py
```diff
@@ -799,3 +833,61 @@ def test_context_as_json_string(self, benchmark, medium_policy, medium_entities)
...
+class PolicySetReuseBenchmarkTestCase:
+    """Benchmarks demonstrating the reusable PolicySet handle across policy sizes.
+
+    Each size has a pair: the string path (which re-parses the policies on every
+    call) vs. a pre-parsed PolicySet handle (which skips the parse). The gap is
+    the per-call policy-parse cost the handle eliminates; it grows with policy
+    size, so the typical (production-scale) pair shows the largest speedup.
+
+    Feature from https://github.com/k9securityio/cedar-py/issues/83
+    """
```
skuenzli: I think the approach of testing both reparsed and reused policyset pairs makes sense.

But I'm curious about the need for a dedicated test suite with a new 'typical' policy. Were the simple/medium/complex policies in PolicyFixtures insufficient to benchmark the PolicySet caching?

tests/unit/test_policyset.py
```diff
@@ -0,0 +1,305 @@
...
+
+        def run(case):
+            request, expected = case
+            return is_authorized(request, ps, ENTITIES).decision == expected
+
+        with ThreadPoolExecutor(max_workers=8) as pool:
+            results = list(pool.map(run, cases))
+
+        self.assertTrue(all(results))
+        self.assertEqual(len(cases), len(results))
```
skuenzli, line 274: very nice. thank you for testing handling of a handle shared across threads

tests/benchmark/test_benchmark_authorize.py
skuenzli: Asking primarily because the cost of maintaining each benchmark test is relatively high, not because I think the original policies are better.

skuenzli (commented):
This PR looks great! Thank you!

I just have the one question in the benchmarking.

tests/benchmark/test_benchmark_authorize.py
skuenzli: Apologies... I think my question is a little confusing.

PolicySetReuseBenchmarkTestCase tests with the following policies
* simple
* medium
* typical

IsAuthorizedBenchmarkTestCase tests with the following policies:
* simple
* medium
* complex
* a couple sandbox_b policies

Do you think it makes merge and 'deduplicate' the PolicySetReuseBenchmarkTestCase tests into  IsAuthorizedBenchmarkTestCase? testing parse + reuse for the following policies:
* simple
* medium
* complex
* typical

Iamrodos:
@skuenzli 

Good call on the testing, done in 051bb24. I've removed the standalone `PolicySetReuseBenchmarkTestCase` and folded the parse-vs-reuse pairs into `IsAuthorizedBenchmarkTestCase`, so each policy size now has a string-path (re-parse) benchmark and a `*_reuse_handle` benchmark right next to it: **simple, medium, complex, large**.

On your other question — the existing `simple`/`medium`/`complex` fixtures are now reused as the re-parse baselines, so the only fixture I kept is the big one, renamed `typical` to `large` to sit at the top of the `simple → medium → complex → large` ladder. I kept it because `complex` is only ~2.9 KB / 10 rules, where reuse is a modest win; the ~9× speedup that motivated #83 only shows at larger scale (`large` is ~16 KB / 60 rules), and none of the existing fixtures reach that size. Happy to drop it if you'd rather not carry a fourth fixture, but it's the one data point that demonstrates the feature actually paying off.

Net change is one added fixture instead of a whole new test class.

skuenzli:
Perfect - thank you!

I also like the renaming of 'typical' to 'large' because it fits the existing pattern and doesn't try to assign what 'typical' or 'normal' is.

The benchmarks are passing:
```
Benchmark                            Baseline (μs)  Median (μs)   Δ median  Status
-----------------------------------  -------------  -----------  ---------  ------
test_batch_complex_policy                     1234         1254      +1.6%  PASS
test_batch_simple_policy                       384          395      +2.7%  PASS
test_batch_size_scaling[100]                  4112         4106      -0.1%  PASS
test_batch_size_scaling[10]                   1105         1119      +1.2%  PASS
test_batch_size_scaling[1]                     799          835      +4.5%  PASS
test_batch_size_scaling[25]                   1595         1617      +1.4%  PASS
test_batch_size_scaling[50]                   2430         2429      -0.0%  PASS
test_batch_size_scaling[5]                     933          965      +3.4%  PASS
test_complex_policy                            273          282      +3.1%  PASS
test_context_as_dict                           802          821      +2.3%  PASS
test_context_as_json_string                    799          821      +2.7%  PASS
test_entities_as_json_string                   727          750      +3.1%  PASS
test_entities_as_list                          796          823      +3.4%  PASS
test_large_entity_set                         3875         3880      +0.1%  PASS
test_medium_entity_set                         795          818      +3.0%  PASS
test_medium_policy                             187          191      +2.1%  PASS
test_sandbox_b_batch_mixed_actions            1098         1117      +1.6%  PASS
test_sandbox_b_batch_multiple_users            683          689      +0.9%  PASS
test_sandbox_b_delete_with_auth                574          585      +1.9%  PASS
test_sandbox_b_view_own_photo                  574          581      +1.1%  PASS
test_sandbox_b_view_public_photo               576          579      +0.6%  PASS
test_simple_policy_allow                       144          148      +3.0%  PASS
test_simple_policy_deny                        143          147      +2.6%  PASS
test_small_entity_set                          187          194      +3.6%  PASS
test_ten_batch_call                           1098         1115      +1.5%  PASS
test_ten_single_calls                         8243         8217      -0.3%  PASS

warnings:
  test_complex_policy_reuse_handle: present in current runs but no baseline — skipped
  test_large_policy: present in current runs but no baseline — skipped
  test_large_policy_reuse_handle: present in current runs but no baseline — skipped
  test_medium_policy_reuse_handle: present in current runs but no baseline — skipped
  test_simple_policy_reuse_handle: present in current runs but no baseline — skipped

N=5 current runs vs tests/benchmark/results/baseline.json; threshold: median Δ > 5% fails
RESULT: PASS
```

I'm also reproducing the headline `PolicySet` reuse claim:
> The benchmarks run clean and reproduce the headline result: test_large_policy (string re-parse) median ~1.51 ms vs test_large_policy_reuse_handle ~157 µs — a ~9.6× speedup, matching the documented ~9.2×.

I'm excited to try this out. 🙌

PR #95, opened 2026-06-23
Entities handle + PolicySet.with_added_str to avoid re-parsing a stable base every call 

Iamrodos:
## Summary

Extends the reusable-handle pattern from policies (#91) to **entities**, and adds a symmetric "add to an existing handle" method on each surface. Both handles now have the same shape: a `from_*` constructor that parses once into an opaque, immutable handle, plus an `add_*` that takes an existing handle and returns a **new** handle with extra content merged in (handles are never mutated in place).

- `Entities.from_json_str(json, schema=None) -> Entities` — parse + transitive closure once
- `entities.add_from_json_str(delta, schema=None) -> Entities` — base plus delta, merge done in Rust, returns a new handle
- `PolicySet.with_added_str(fragment) -> PolicySet` — clone the compiled base, parse only the fragment, returns a new handle

An `Entities` handle is now accepted anywhere the `entities` argument is accepted today — `is_authorized`, `is_authorized_batch`, `is_authorized_partial` — exactly as a `PolicySet` is already accepted for `policies`.

Closes the entities half of #83 (the `PolicySet` handle was the policies half, #91).

## Motivation

Profiling a downstream service showed the same per-call waste #83 found for policies, but for entities: it builds a large, stable base entity graph (a few hundred entities) once, then runs many authorization checks, each adding only a tiny per-request delta (the principal plus a couple of resource entities). Today every `is_authorized` call re-serialises the whole graph to JSON in Python and the Rust core re-parses it — re-deserialising and recomputing the transitive closure each time. A reusable `Entities` handle lets the parsed base be reused, with only the delta parsed per call. `PolicySet.with_added_str` closes the matching gap for policies: a caller with a static base and small dynamic fragments no longer has to re-parse the whole base text each call.

## Usage

```python
from cedarpy import Entities, PolicySet, is_authorized

# Stable base graph (org users/groups): parse once, reuse for every request.
base = Entities.from_json_str(org_graph_json)

for request, request_entities_json in workload:
    # Parse only this request's entities; the base is not re-parsed.
    for_request = base.add_from_json_str(request_entities_json)
    is_authorized(request, policy_set, for_request)
```

```python
# Mostly-static policies plus a per-request fragment:
base = PolicySet.from_str(static_policies)
for_request = base.with_added_str(per_request_policies)
is_authorized(request, for_request, entities)
```

A handle is accepted anywhere the corresponding string/list is accepted; passing a string or list still works exactly as before. The handles compose — reuse whichever inputs are stable for your workload, or both.

## Design decisions

- **Mirror the `PolicySet` handle precedent (#91) exactly.** Same opaque `frozen` `#[pyclass]`, same `str | handle` (`FromPyObject`) extractor with the string variant tried first (so a `str`/`list` caller pays no failed handle extraction), same borrow-into-slot on the authz path, same eager-`ValueError`-on-construction.
- **Handles are immutable; `add_*` returns a NEW handle.** cedar's `Entities::add_entities*` / `PolicySet::merge` consume/mutate, but cedarpy clones the base first so a single compiled base can be shared across threads and requests without a clone-before-mutate footgun.
- **`with_added_str` stays equivalent to concatenated text.** cedar derives a surface-syntax `PolicyId` (`policy0`, `policy1`, …) positionally **per parse**, so a fragment parsed on its own restarts at `policy0` and would collide with the base. The fragment's colliding ids are renumbered to follow the base (via cedar's `merge(rename_duplicates=true)`, which also updates internal references), exactly as parsing the concatenated text would number them. **`@id` annotations are preserved** and still resolve via `diagnostics.id_annotations_by_reason` — only the positional `PolicyId` is adjusted. (There is a test pinning this, given the repo's `@id` history in #66/#68/#74/#77.)
- **Backward compatible.** The `entities` parameter widens from `str | list` to `str | list | Entities`; the existing string and list forms are unchanged. Same for `policies` (already `str | PolicySet`).

## Behavior changes

| Change | Before | After |
|--------|--------|-------|
| **metrics** | — | successful results gain `entities_pre_parsed` (`1` when an `Entities` handle is passed, `0` for a string/list). Mirrors the existing `policies_pre_parsed`. |

Additive / opt-in for string callers.

## Performance

Release build, median `is_authorized`, via `pytest tests/benchmark --benchmark-only -k "entity_set or base_plus_delta"`:

| Scenario | string (re-parse) | handle (reuse) | speedup |
|--------|-------------------|----------------|---------|
| medium entity set (~50 entities) | ~799 µs | ~99 µs | ~8× |
| large entity set (~200 entities) | ~3.84 ms | ~106 µs | **~36×** |
| stable base (~200) + per-request delta | ~3.86 ms | ~243 µs | **~16×** |

The win scales with the size of the **stable base**: reuse removes the per-call JSON deserialization of the base. The base-plus-delta path (~243 µs) is honestly slower than pure reuse (~106 µs) because it still parses the delta and recomputes the transitive closure each call — what it avoids is re-deserialising the ~200-entity base. (Schema, like before, is still parsed each call.)

## Test plan

- [x] `maturin develop --release` — clean build, no warnings
- [x] `pytest tests/unit` — **156 passed** (was 126; +30 new across `tests/unit/test_entities.py` and `tests/unit/test_policyset.py`: round-trip, string/list/handle equivalence across `is_authorized` / `_batch` / `_partial`, the realistic base-plus-delta scenario, immutability, duplicate-uid + identical-dup + duplicate-policy-id behavior, malformed-JSON `ValueError`, schema compose/reject, two-independent-handles, shared-handle-across-threads with per-thread deltas, wrong-type `TypeError`, `@id` survives `with_added_str` renumbering, the new metric)
- [x] `make benchmark-compare` — **performance-neutral.** A direct HEAD-vs-`main` A/B on the affected hot-path benchmarks is within run-to-run noise (`main`'s own `test_batch_simple_policy` swings ~6% between back-to-back runs in this session; the change's reading sits inside that band). The string entities path was also made allocation-free (parses `&str` directly instead of cloning). The gate reports one benchmark (`test_batch_simple_policy`) over its 5% threshold against `baseline.json`, but that baseline was captured at 4.8.0 and predates ~20 merged commits, so it reflects accumulated drift + that benchmark's noise, not this change — same situation noted in #91. New reuse benchmarks have no baseline and are skipped.
- [x] `cedarpy/_internal.pyi` updated; `pyright` clean for the changed surface


cedarpy/_internal.pyi
```diff
@@ -34,6 +34,81 @@ class PolicySet:
...
+    def from_json_str(s: str, schema: Optional[str] = ...) -> "Entities":
+        """Parse an ``Entities`` handle from a Cedar JSON entities document.
+
+        ``schema`` (optional) is Cedar schema text or JSON; when supplied, the
+        entities are validated against it. Raises ``ValueError`` if the entities
+        (or schema) cannot be parsed, or the entities violate ``schema``.
+        """
+        ...
+
+    def add_from_json_str(self, delta: str, schema: Optional[str] = ...) -> "Entities":
```
skuenzli: Should `Entities#add_from_json_str` be named `Entities#with_added_json_str` to parallel `PolicySet#with_added_str`?

Seems like we can align one way or another and `with_` communicates the method will return a new object.

src/lib.rs
```diff
@@ -110,6 +110,39 @@ impl PyPolicySet {
...
+        // restart at `policy0`) composes with a non-empty base exactly as the
+        // concatenated text would. cedar updates any internal references to the
+        // renamed ids. The only error path left is a malformed fragment, caught
+        // by `from_str` above; `merge` itself cannot fail here.
+        merged
+            .merge(&added, true)
+            .map_err(|e| pyo3::exceptions::PyValueError::new_err(e.to_string()))?;
+        Ok(PyPolicySet { inner: merged })
+    }
+
```
skuenzli, line 145: 👍

cedarpy/_internal.pyi
Iamrodos: Good call, let’s align on `with_`. I’ll rename `add_from_json_str` to `with_added_json_str` so it parallels`PolicySet.with_added_str` and signals that it returns a new handle rather than mutating in place. Keeping the `_json_str` part, since entities only have a JSON form and it matches the `from_json_str` constructor. I’ll update the stub, the Rust method, the tests, and the docs in this PR.

src/lib.rs
```diff
@@ -125,6 +158,138 @@ impl PyPolicySet {
...
+        match self.inner.write_to_json(&mut buf) {
+            Ok(()) => String::from_utf8_lossy(&buf).into_owned(),
+            Err(e) => format!("<unrenderable Entities: {e}>"),
+        }
+    }
+
+    fn __repr__(&self) -> String {
+        format!("Entities(<{} entities>)", self.inner.iter().count())
+    }
+}
```
skuenzli, line 261: PyEntities looking good

cedarpy/_internal.pyi
Iamrodos: Done.

skuenzli: thank you.

tests/unit/test_entities.py
```diff
@@ -0,0 +1,405 @@
...
+        self.assertEqual(from_str.diagnostics.errors, from_handle.diagnostics.errors)
+
+    def test_owner_allowed_matches_string_path(self) -> None:
+        self.assertEqual(Decision.Allow, is_authorized(ALICE_VIEW, OWNER_POLICY, self.entities).decision)
+        self._assert_equivalent(ALICE_VIEW)
+
+    def test_non_owner_denied_matches_string_path(self) -> None:
+        self.assertEqual(Decision.Deny, is_authorized(BOB_VIEW, OWNER_POLICY, self.entities).decision)
+        self._assert_equivalent(BOB_VIEW)
+
```
skuenzli, line 107: 👍

tests/unit/test_entities.py
```diff
@@ -0,0 +1,405 @@
...
+    def test_str_list_and_handle_all_agree(self) -> None:
+        entities_list = json.loads(ENTITIES_JSON)
+        handle = Entities.from_json_str(ENTITIES_JSON)
+        from_str = is_authorized(ALICE_VIEW, OWNER_POLICY, ENTITIES_JSON).decision
+        from_list = is_authorized(ALICE_VIEW, OWNER_POLICY, entities_list).decision
+        from_handle = is_authorized(ALICE_VIEW, OWNER_POLICY, handle).decision
+        self.assertEqual(Decision.Allow, from_str)
+        self.assertEqual(from_str, from_list)
+        self.assertEqual(from_str, from_handle)
+
```
skuenzli, line 122: 👍

tests/unit/test_policyset.py
```diff
@@ -274,6 +274,73 @@ def run(case):
...
+    def test_result_handle_usable_in_batch_and_partial(self) -> None:
+        combined = PolicySet.from_str(self.BASE).with_added_str(self.FRAGMENT)
+        alice_req = {"principal": 'User::"alice"', "action": 'Action::"view"', "resource": 'Photo::"p"'}
+        bob_req = {"principal": 'User::"bob"', "action": 'Action::"view"', "resource": 'Photo::"p"'}
+        batch = is_authorized_batch([alice_req, bob_req], combined, ENTITIES)
+        self.assertEqual([Decision.Allow, Decision.Allow], [r.decision for r in batch])
+        partial = is_authorized_partial({**alice_req, "context": {}}, combined, ENTITIES)
+        self.assertEqual(Decision.Allow, partial.decision)
+
+
```
skuenzli, line 343: 👍

skuenzli (approved):
This looks great. Thank you!

I think the PolicySet and Entities handles open up simpler, more literate, authorization implementations for consumers.

PR #96, opened 2026-06-24
feat: link Cedar policy templates on the PolicySet handle

### LINKED ISSUE #29: Support for policy templates

Iamrodos:
## Summary

Surfaces Cedar **policy templates** (policies with `?principal` / `?resource` slots) on the reusable `PolicySet` handle, extending the handle pattern from #91/#95 to template linking. `PolicySet.from_str` already parses templates; they just authorize nothing until linked. This adds the full linking lifecycle, immutable like the rest of the handle — every method returns a **new** `PolicySet`:

- `policy_set.with_linked_batch(links) -> PolicySet` — the primary path: clone the base once, apply every link, return a new handle. All-or-nothing (any failure raises, no handle returned).
- `policy_set.with_linked(template_id, new_id, values) -> PolicySet` — single-link convenience over the batch path.
- `policy_set.templates() -> list[dict]` — introspection: one dict per template with its `id`, `id_annotation` (the `@id`), `slots`, and the `links` derived from it (each link's `id` and bound `values`).
- `policy_set.without_linked(link_id) -> PolicySet` — a new handle with one linked policy removed.

A linked `PolicySet` is still just a `PolicySet`, so it evaluates through `is_authorized`, `is_authorized_batch`, and `is_authorized_partial` with no other changes. Existing callers are untouched — pure, opt-in addition.

Implements the template-linked policy support discussed in #29. (Happy to close #29 on merge, or leave it open if you'd rather keep tracking the other items from the thread — your call.)

> ### 📖 Please read the README "Linking policy templates" section first
> The [README section](README.md#linking-policy-templates) is the canonical reference for this change and intentionally covers a lot: the linking mechanics, the **`@id`-as-stable-key** reasoning (and why the positional `policy0` id is the fragile one), what a linked policy's `id` vs `@id` are, the `templates()` shape, and an honest **"should you reach for templates?"** efficiency discussion. The design-decision and performance notes below summarize it, but the README is where the reasoning lives.

## Motivation

Templates are Cedar's first-class, GA mechanism for **per-principal and per-resource grants** — "allow this person to view this photo while their subscription is active" — authored once and linked per grant, instead of an N-near-identical-policies string-concatenation. Two concrete consumers drove the shape (see #29): per-user permission grants in my own app, and the canonical per-principal sharing / multi-tenancy pattern templates were designed for. The crate already exposes everything needed (`PolicySet::link` / `unlink` / `templates` / `get_linked_policies`); the gap was a Python surface, which this fills following the immutable-handle precedent.

## Usage

```python
from cedarpy import PolicySet, is_authorized, Decision

# one rule: a subscriber may view a photo that's been granted to them
base = PolicySet.from_str(
    '@id("photo-access")\n'
    'permit(principal == ?principal, action == Action::"view", resource == ?resource)\n'
    'when { principal.subscriptionActive };'
)

# fill the slots to grant alice access to one photo
linked = base.with_linked(
    template_id="photo-access",
    new_id="alice-vacation",
    values={"?principal": 'User::"alice"', "?resource": 'Photo::"vacation.jpg"'},
)

entities = [{"uid": {"type": "User", "id": "alice"},
             "attrs": {"subscriptionActive": True}, "parents": []}]
request = {"principal": 'User::"alice"', "action": 'Action::"view"', "resource": 'Photo::"vacation.jpg"'}

result = is_authorized(request, linked, entities)
assert result.decision == Decision.Allow
result.diagnostics.reasons                    # ['alice-vacation']  (the link's id)
result.diagnostics.id_annotations_by_reason   # {'alice-vacation': 'photo-access'}  (inherited @id)
```

```python
# grant many at once — with_linked_batch clones the base once and applies every link
linked = base.with_linked_batch([
    {"template_id": "photo-access", "new_id": "alice-vacation",
     "values": {"?principal": 'User::"alice"', "?resource": 'Photo::"vacation.jpg"'}},
    {"template_id": "photo-access", "new_id": "bob-skyline",
     "values": {"?principal": {"type": "User", "id": "bob"}, "?resource": 'Photo::"skyline.jpg"'}},
])
```

Slot values accept a Cedar string (`'User::"jane"'`) or a `{"type","id"}` dict — the same two forms a request principal/resource already accepts (#87).

## Design decisions

- **Mirror the immutable-handle precedent (#91/#95).** Cedar's `link`/`unlink` mutate a `PolicySet`; cedarpy clones the base and returns a **new** handle, so a single compiled base is shared across threads/requests without a clone-before-mutate footgun. Every mutating-sounding method (`with_linked`, `with_linked_batch`, `without_linked`) returns a new handle.
- **`with_linked_batch` is primary; `with_linked` is sugar.** Per the #29 discussion, the batch path is the real primitive: it clones the base **once** and applies all links (one clone per batch, not per link), and it's all-or-nothing. `with_linked` delegates to it. The `_batch` suffix mirrors the repo's own `is_authorized` / `is_authorized_batch` convention.
- **Identify templates by `@id`, not the positional id.** A `template_id` resolves to a template's literal Cedar id first, then to its `@id` annotation value (the labeling convention the Cedar CLI's `link` uses; `@id` is otherwise inert, per the repo's #66/#68/#74/#77 history). **Prefer the `@id`:** the positional id (`policy0`, …) is reassigned per parse and renumbered by `with_added_str`, while the `@id` is invariant. Resolution never renames or rebuilds the set, and the `@id` match must be unambiguous. Resolution is **indexed once per batch** (`build_template_id_index`), so a batch of `L` links over `T` templates is `O(T + L)`, not `O(T × L)`; the raw-`&str` annotation scan avoids Cedar's identifier-parse cost (same pattern as `lookup_id_annotation`).
- **`new_id` matches Cedar's naming across surfaces.** The link-spec keys `template_id` / `new_id` / `values` are the snake_case of Cedar's own: Rust `PolicySet::link(template_id, new_id, vals)`, CLI `--template-id` / `--new-id`, and the EST/JSON `templateId` / `newId` / `values`. `new_id` is the id of the new template-linked policy (the filled-in slots, which evaluate as a policy) — neither a template id nor a principal.
- **`templates()` is the one introspection call.** It returns the template definition *and* the links nested under each template (each link's `id` and bound `values`), composed from `templates()` / `Template::slots()`/`annotations()` / `get_linked_policies` / `template_links()`. A linked policy's id is its `new_id` (what `reasons` reports); its `@id` is inherited from the template.
- **Refactor: shared entity-uid extraction.** `extract_euid_field` (request principals) and slot-value parsing now share `euid_input_from_value`, with the error label computed lazily (a closure) so the hot success path allocates no label — request error messages are byte-identical.

## Behavior changes

None for existing callers. Purely additive: a `PolicySet` may now contain template-linked policies, and the new methods are opt-in. No new metrics, no new Cargo feature — regular template linking works on the pinned engine (4.8.2).

## Performance

**Authorization hot path: neutral.** A linked `PolicySet` evaluates exactly like a static one; the only `is_authorized`-path change is the `euid_input_from_value` refactor, which is allocation-free on the success path.

**Build cost (the interesting axis), and an honest one — this is *not* a general speedup.** A quick build benchmark (not committed; see "Should you reach for templates?" in the README) of generating + parsing N policies vs one template + N links via `with_linked_batch`, at N=200 grants:

| policy body | build speed | memory | winner |
|---|---|---|---|
| 1 condition (subscription) | 0.7× | 1.0× | generate+parse |
| ~4 conditions | 1.2× | 1.8× | ~parity |
| ~10 conditions (realistic rich) | **~2×** | **~2.5×** | templates |

(speed/mem = generate+parse ÷ template; >1 favors templates.) Templates pay off as the policy body gets heavier — the rule is parsed and held *once* and shared across links instead of copied per grant — and are slightly *worse* for trivial policies (the per-link binding exceeds a tiny static policy). For typical/basic policies it's roughly a wash; the README says so plainly and tells readers to profile their own shapes.

## Test plan

- [x] `maturin develop --release` — clean build, no warnings
- [x] `pytest tests/unit` — **198 passed** (+42 new in `tests/unit/test_template_linking.py`: parse/round-trip, single + batch link evaluation through `is_authorized`/`_batch`/`_partial`, immutability, batch clone-once + all-or-nothing atomicity (incl. intra-batch duplicate `new_id`), both slot-value forms, `?principal`/`?resource`/both slots, `@id`-vs-literal resolution + ambiguity, `templates()` shape incl. nested links, `without_linked` + relink, schema compose, shared-handle-across-threads, and an **upstream-anchored round-trip over Cedar's own `sandbox_c` sample**, vendored under `tests/unit/resources/sandbox_c/` with provenance)
- [x] `pyright` clean for the changed surface (`cedarpy/_internal.pyi` updated)
- [x] `make benchmark-compare` — **performance-neutral.** The gate reports `test_simple_policy_allow`/`_deny` over the 5% threshold (+5.5% / +5.6%), but this is `baseline.json` drift + warm-machine noise, not this change: the whole suite is uniformly elevated (+1–5.5%, including benchmarks this PR doesn't touch, e.g. `test_small_entity_set` +4.8%), the two flagged benchmarks are the trivial-policy authz path (unchanged here bar the allocation-free `euid_input_from_value` refactor), and a HEAD-vs-`main` A/B confirmed `main` itself trips the gate on this machine — the same `baseline.json`-predates-many-merged-commits drift noted in #91/#95. New template benchmarks have no baseline and are skipped.

## Questions for review

- **Tracked benchmark?** The README cites build-time + memory numbers from a quick ad-hoc benchmark (worded as such, not committed). Happy to add a committed, reproducible template build+memory benchmark under `tests/benchmark/` if you'd like — it'd be a standalone script, *not* through the median-of-N authz regression gate (it's build/memory, not authz timing).
- **Leveraging upstream test data:** this vendors Cedar's `sandbox_c` and reproduces its documented `link` walkthrough; happy to pull in more cedar-policy / cedar-java template fixtures if you want broader coverage (per your note in #29).

---


skuenzli:
@Iamrodos - thank you for the PR!

My initial review of the change is that it looks good. But I want to review it in a bit more detail tomorrow. Thanks for your patience.

README.md
```diff
@@ -173,6 +173,86 @@ for request in requests:
...
+
+**Identify a template by its `@id`, not its positional id.** Cedar assigns a surface-syntax policy a positional id — `policy0`, `policy1`, … — per parse, and that position can shift (`with_added_str`, for example, renumbers an incoming fragment's ids when it layers them onto a base). A template's `@id` annotation is invariant across parsing and merging, so it is the stable key to link against. `with_linked` accepts either — it matches the literal id first, then the `@id` — but prefer the `@id`. The match must be unambiguous; two templates sharing one `@id` raises rather than guessing. (An `@id` is [otherwise inert](https://docs.cedarpolicy.com/policies/syntax-policy.html#term-parc-annotations) and is not the template's id; cedarpy resolves it the way the Cedar CLI's `link` command does.)
```
skuenzli: The first sentence needs some work. I am having trouble parsing it, ironically.

Perhaps:
> Cedar assigns a policy id by position, `policy0`, `policy1`..., per parse. That position can shift ...

README.md
```diff
@@ -173,6 +173,86 @@ for request in requests:
...
+assert result.decision == Decision.Allow
+
+# the decision points at the link's own id; the template's @id comes alongside it
+result.diagnostics.reasons                    # ['alice-vacation']
+result.diagnostics.id_annotations_by_reason   # {'alice-vacation': 'photo-access'}
+```
+
+Linking produces a *template-linked policy* — the slots filled with concrete values, which evaluates as a policy in its own right — and `new_id` is the id assigned to it. Linking does not rename or consume the template: the template stays in the set, so you can link it again under a different `new_id` to grant another principal (that's the whole point — one template, many grants). `new_id` is yours to choose; it is neither a template id nor a principal. (This mirrors the Cedar CLI's `link --template-id … --new-id …`.)
+
+To grant many principals at once, use `with_linked_batch`. It is the primary linking path: it clones the base **once** and applies every link to that clone, so a batch pays a single clone rather than one per link, and it is all-or-nothing (if any link fails the whole call raises and no handle is returned):
```
skuenzli: I don't think 'To grant many _principals_ at once' is correct (emphasis added) because templating supports both principals and resources. Plus, you can specify different templates.

So perhaps:

> To link many templates at once, use `with_linked_batch`.

or 
> To create many template-linked policies at once, use `with_linked_batch`.

Then... what do you think about explaining what it does for the user:
> This is the primary linking path and enables efficient linking of any template in the `PolicySet` to the entities that you choose.

Of the implementation details in the current paragraph, I think it is most important to communicate that the operation is all or nothing.

README.md
```diff
@@ -173,6 +173,86 @@ for request in requests:
 
 > Note: the `Entities` and `PolicySet` handles are independent and compose — reuse whichever inputs are stable for your workload, or both. Schema is still parsed on each call.
 
+### Linking policy templates
+
+A Cedar [policy template](https://docs.cedarpolicy.com/policies/templates.html) is a policy with `?principal` / `?resource` *slots* — a rule written once and then *linked* to concrete entities to produce real, evaluatable policies. The canonical case is per-principal and per-resource grants — *allow this person to view this photo while their subscription is active*. You write that rule once as a template, then link it per grant.
```
skuenzli: This is the first use of the term/jargon `slot`. I think we should define it.

Perhaps:

> A 'slot' is a placeholder in a policy template that marks where a principal or resource can be populated later through the linking process.

README.md
```diff
@@ -173,6 +173,86 @@ for request in requests:
...
+    'permit(principal == ?principal, action == Action::"view", resource == ?resource)\n'
+    'when { principal.subscriptionActive };'
+)
+
+# fill the slots to grant alice access to one photo
+linked = base.with_linked(
+    template_id="photo-access",          # which template to fill in
+    new_id="alice-vacation",             # id for this link (the filled slots, which act as a policy)
+    values={"?principal": 'User::"alice"', "?resource": 'Photo::"vacation.jpg"'},
+)
```
skuenzli, line 211: this is a great intro example

tests/unit/test_template_linking.py
```diff
@@ -0,0 +1,517 @@
...
+        linked = _link_alice(PolicySet.from_str(PRINCIPAL_TEMPLATE))
+        static = PolicySet.from_str(
+            'permit(principal == User::"alice", action == Action::"view", resource);'
+        )
+        for req in (ALICE_VIEW, BOB_VIEW):
+            self.assertEqual(
+                is_authorized(req, static, ENTITIES).decision,
+                is_authorized(req, linked, ENTITIES).decision,
+            )
+
```
skuenzli, line 155: 👍

tests/unit/test_template_linking.py
```diff
@@ -0,0 +1,517 @@
...
+        linked = base.with_linked("policy0", "by-literal", {"?principal": 'User::"alice"'})
+        self.assertEqual(Decision.Allow, is_authorized(ALICE_VIEW, linked, ENTITIES).decision)
+
+    def test_linking_creates_new_policy_without_consuming_template(self) -> None:
+        # new_id names a NEW policy; the template is not renamed or consumed and
+        # can be linked again under a different new_id (one template, many grants)
+        base = PolicySet.from_str(PRINCIPAL_TEMPLATE)
+        once = base.with_linked("grant-view", "grant-for-alice", {"?principal": 'User::"alice"'})
+        twice = once.with_linked("grant-view", "grant-for-bob", {"?principal": 'User::"bob"'})
+        self.assertEqual(["grant-for-alice", "grant-for-bob"], link_ids(twice))
```
skuenzli, line 139: nice

tests/unit/test_template_linking.py
```diff
@@ -0,0 +1,517 @@
...
+    """with_linked_batch is the primary path; single linking is sugar over it."""
+
+    def test_batch_links_many_in_one_call(self) -> None:
+        base = PolicySet.from_str(PRINCIPAL_TEMPLATE)
+        linked = base.with_linked_batch([
+            {"template_id": "grant-view", "new_id": "alice",
+             "values": {"?principal": 'User::"alice"'}},
+            {"template_id": "grant-view", "new_id": "bob",
+             "values": {"?principal": 'User::"bob"'}},
+        ])
```
skuenzli, line 174: if I was going to add one thing to the test cases, it would be demonstrating that with_linked_batch supports linking to different templates, e.g. `grant-view` and `grant-edit`

skuenzli (changes requested):
@Iamrodos - I think the code is good and the docs are nearly ready. I suggested a few changes to the README that I think will make it easier for new users to consume.

Thanks for building this!

Iamrodos:
>  I suggested a few changes to the README that I think will make it easier for new users to consume.

@skuenzli changes done to README and one test

Thanks for the good review as always.

tests/unit/test_template_linking.py
skuenzli: 🙌

README.md
```diff
@@ -175,19 +175,33 @@ for request in requests:
 
 ### Linking policy templates
 
-A Cedar [policy template](https://docs.cedarpolicy.com/policies/templates.html) is a policy with `?principal` / `?resource` *slots* — a rule written once and then *linked* to concrete entities to produce real, evaluatable policies. The canonical case is per-principal and per-resource grants — *allow this person to view this photo while their subscription is active*. You write that rule once as a template, then link it per grant.
+A Cedar [policy template](https://docs.cedarpolicy.com/policies/templates.html) is a policy with `?principal` / `?resource` *slots*. A **slot** is a placeholder that marks where a principal or resource is filled in later, when the template is *linked* to concrete entities to produce a real, evaluatable policy. So a template is a rule written once and linked per use. The canonical case is per-principal and per-resource grants — *allow this person to view this photo while their subscription is active*. You write that rule once as a template, then link it per grant.
```
skuenzli, line 178: terrific - the language is consistent with [cedar policy template docs](https://docs.cedarpolicy.com/policies/templates.html) too

README.md
```diff
@@ -211,20 +225,42 @@ result.diagnostics.id_annotations_by_reason   # {'alice-vacation': 'photo-access
...
 
-To grant many principals at once, use `with_linked_batch`. It is the primary linking path: it clones the base **once** and applies every link to that clone, so a batch pays a single clone rather than one per link, and it is all-or-nothing (if any link fails the whole call raises and no handle is returned):
+To create many template-linked policies at once, use `with_linked_batch`. This is the primary linking path: in a single call it fills the slots of any templates in the `PolicySet` — the same one repeatedly or several different ones — with the entities you choose. If any link in the batch fails, the whole call raises and no handle is returned. It clones the base **once** rather than per link:
```
skuenzli, line 228: very nice

skuenzli (approved):
Looks great. Thank you!

PR #97, opened 2026-06-26
fix(templates): return slots sorted, document as a set

skuenzli:
## Summary

Follow-up to #96. `PolicySet.templates()` returned each template's `slots` as a `list` in `Template::slots()`'s **unspecified** iteration order — so a two-slot template came back as `['?resource', '?principal']` at runtime, which didn't match the README example's `['?principal', '?resource']` (the example wasn't reproducible).

A template's slots are conceptually a **set**: each of `?principal` / `?resource` appears at most once (a link's `values` is keyed by slot), so order carries no meaning.

## Change

- **Sort the slot keys** in `templates()` for a deterministic, reproducible result (`src/lib.rs`). Since `"?principal" < "?resource"` lexically, the sorted output also matches the existing README example, so the docs become accurate without changing the example.
- **Document `slots` as a set** whose order is not part of the contract — in the Rust docstring, the `cedarpy/_internal.pyi` stub, and the README.
- **Test** (`test_slots_is_a_set_returned_deterministically`): pins the set contract (membership `{?principal, ?resource}`, no duplicates) and the deterministic sorted order.

`slots` stays a JSON-serializable `list` (so the whole `templates()` payload remains `json.dumps`-able) — this just makes it deterministic and documents the set semantics.

## Verification

- `maturin develop --release` — clean
- `pytest tests/unit` — **199 passed** (+1 new); `make integration-tests` — 74 passed
- Runtime `templates()[i]["slots"]` is now `['?principal', '?resource']`, matching the README example

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ revoked = linked.without_linked("bob-skyline")
 # revoked.templates()[0]['links'] now lists only alice-vacation
 ```
 
-`templates()` is the one introspection call you need to understand template declarations and links. `tempaltes` returns each template's `id`, `id_annotation`, and `slots` declarations plus the `links` produced from it (each link's `id` and bound `values`). To act on a single grant, e.g. revoke it, read its id from `links` and pass it to `without_linked(link_id)`.
+`templates()` is the one introspection call you need to understand template declarations and links. `templates` returns each template's `id`, `id_annotation`, and `slots` declarations plus the `links` produced from it (each link's `id` and bound `values`). To act on a single grant, e.g. revoke it, read its id from `links` and pass it to `without_linked(link_id)`.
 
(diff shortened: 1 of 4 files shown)


skuenzli:
@Iamrodos - what do you think of this change? Smoothing out a nit.

Iamrodos:
Your right, it is a nit, but fair.

Note that the rewrite drops the old "(Order within templates and links is not guaranteed.)". Since this PR only sorts slots, those two arrays are still unordered. Might be worth keeping a short version so readers don't assume they're ordered now. Fine if you merge without it though.

skuenzli:
Dropping "(Order within templates and links is not guaranteed.)" was intentional. 

I try not to discuss non-contract behavior, especially in the README.

I also considered (quietly) ordering the templates and links. Do you think we should do that for consistency?

Iamrodos:
> I also considered (quietly) ordering the templates and links. Do you think we should do that for consistency?

Personally I would not sort any of them by default, including slots. It doesn’t feel pythonic. These structures (dict, list, set) aren’t sorted by default; Python preserves insertion order rather than imposing one. An example showing the output in a certain order in a list is for readability while still being completely valid.

skuenzli:
Ok - thanks for your feedback.

I'm going to close this.

Have a great weekend!

Issue #85, opened 2026-05-28
Clean up stale windows-sys 0.48 transitive deps in Cargo.lock


skuenzli:
## Summary

`Cargo.lock` carries a duplicate `windows-sys 0.48` family alongside the live `0.59` family. The 0.48 copies are reachable only through a Windows-only `[build-dependencies]` chain, but they sit in the lock as dead duplicates and can show up in `cargo audit` / SBOM output.

## Analysis

The 0.48 entries on main's lock are pulled by:

```
windows-sys 0.48.0 → winapi-util → walkdir → lalrpop [build-dependencies] → cedar-policy-core
```

Key properties of this chain:

- **Build-time only**: `lalrpop` is a parser generator that runs during `cargo build` to produce parser tables for `cedar-policy-core`. None of these crates link into the final cdylib.
- **Windows-only**: `winapi-util` only references `windows-sys` under `#[cfg(windows)]`. On macOS/Linux wheel builds, the modules referencing it never compile.
- **Duplicated**: `windows-sys 0.59.0` is already in the lock, pulled by `clap`, `miette`, `rustix`, etc. The 0.48 copy is a parallel resolution that nothing else shares.

The cedarpy wheel itself does not expose any of this to Python consumers:

- `crate-type = ["cdylib"]` — no rlib, not on crates.io, no Rust API surface
- `[build-dependencies]` outputs never enter the shipped artifact
- No FFI passthrough in the Python wrapper

So the entries are safe to remove. Doing so reduces noise in `cargo audit` / SBOM output and removes a potential source of false-positive advisories on a build-only Windows crate line.

## Proposed approach

Either:

1. Bump or adjust `winapi-util` / `walkdir` so the build-deps resolve `windows-sys` to the same `0.59` line already in the lock, or
2. Allow `cargo update` to consolidate naturally and just commit the slimmed lock.

Option 2 was observed locally to drop these 9 entries cleanly. Either way: keep this in its own PR with no source changes, so the lock-only diff is easy to review.

## Affected entries

The 9 lock entries to remove:

- `windows-sys 0.48.0`
- `windows-targets 0.48.5`
- `windows_aarch64_gnullvm 0.48.5`
- `windows_aarch64_msvc 0.48.5`
- `windows_i686_gnu 0.48.5`
- `windows_i686_msvc 0.48.5`
- `windows_x86_64_gnu 0.48.5`
- `windows_x86_64_gnullvm 0.48.5`
[...]

PR #106, opened 2026-08-21
deps: update cedar-policy to 4.12.0

h0rv:
`Cargo.toml` already permits this, since the requirement on `cedar-policy` is a caret on `4.8.2`. Only `Cargo.lock` held the engine back. No source changes were needed.

```
cedar-policy           4.8.2 -> 4.12.0
cedar-policy-cli       4.8.2 -> 4.12.0
cedar-policy-core      4.8.2 -> 4.12.0
cedar-policy-formatter 4.8.2 -> 4.12.0
```

Done with a targeted `cargo update -p cedar-policy -p cedar-policy-cli -p cedar-policy-formatter` rather than a blanket update, per the `Cargo.lock` guidance in `CLAUDE.md`.

`cedar-policy` 4.12.0 was published 2026-07-28, so it is outside the 7 day window in the version-selection policy.

## Test results

Built with `maturin develop --release` on macOS arm64, Python 3.13.

| Suite | Result |
|---|---|
| `pytest tests/unit` | 218 passed, 2 subtests passed |
| `pytest tests/integration/test_cedar_integration_tests.py` | 74 passed |
| `pytest tests/integration/test_cedar_corpus_tests.py` | 59696 passed |

## Lockfile churn

9 crates added, 4 dropped. All of it comes from Cedar's own dependency changes between 4.8.2 and 4.12.0.

Added: `jiff` 0.2.1, `jiff-tzdb` 0.1.8, `jiff-tzdb-platform` 0.1.3, `bs58` 0.5.1, `itertools` 0.15.0, `syn` 3.0.3, `rustc-literal-escaper` 0.0.8, `portable-atomic-util` 0.2.7, `windows-sys` 0.61.2.

Dropped: `beef` 0.5.2, `lazy_static` 1.5.0, `rustc_lexer` 0.1.0, `rustc_version` 0.4.1.

Two of those are worth a look if you want one. The `jiff` family backs Cedar's datetime extension and brings a timezone database with it, which is a real addition to the dependency surface. `windows-sys` lands at 0.61.2, which is relevant given the 0.48 cleanup in #85 and #98.

## Not run

`make benchmark-compare`. The machine I built on was compiling and running suites throughout, and `CLAUDE.md` says the comparison is load-sensitive and that a uniform shift reads as contamination rather than a regression. A number from that machine would not be trustworthy, so I left it for CI or for a quiet machine. Flagging it because an engine bump is the kind of change where the median gate is worth trusting.

## Why I want it

I use partial evaluation to build a SQL pre-filter for list endpoints, and I read policy structure at startup to decide which principal attributes to populate. Both of those are tree walks over the EST JSON today. Cedar 4.11.0 added the public `pst` module, which is a better fit for the second one, and none of it is reachable at 4.8.2. #107 has the details, including what I found does and does not work. This PR is only the bump and does not depend on that discussion going anywhere.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,7 +71,7 @@ version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "79947af37f4177cfead1110013d678905c37501914fba0efea834c3fe9a8d60c"
 dependencies = [
- "windows-sys",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -81,7 +81,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2109dbce0e72be3ec00bed26e6a7479ca384ad226efdd66db8fa2e3a38c83125"
 dependencies = [
  "anstyle",
- "windows-sys",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


skuenzli:
Hi @h0rv - thanks for the PR to update the Cedar Policy engine!

I will take a look.

skuenzli:
Hi @h0rv - I've completed the upgrade to cedar-policy 4.12.0 in PR #112

PR #118, opened 2026-09-19
fix: report the cause when an authorization request cannot be built

david-long1:
A caller of `is_authorized` or `is_authorized_batch` whose request `context` holds a JSON `null` or a float gets `Decision.NoDecision` with the single diagnostic `failed to parse schema from request`. The message names the schema even when `schema=None` was passed and says nothing about the context. The same string covers every failure inside request construction: a principal, action, or resource that does not parse, a context that does not match the action's schema, and a request that fails schema validation. `tests/unit/test_authorize.py` pinned that exact string, so the flattening was baked in rather than noticed.

The cause is in `src/lib.rs`. `execute_authorization_request` wraps the error from `RequestArgs::get_request` with `e.context("failed to parse schema from request")`, and `make_authz_result_for_errors` renders each error through `stringify_errors`, which calls `e.to_string()`. `Display` on an `anyhow::Error` prints only the outermost context, so the chain is dropped before it reaches Python. The `verbose` console path and the policy-parse path in the same file already render with `{:#}` (the alternate form, which prints the chain as `outer: inner`), so only the returned diagnostics lost the cause. `is_authorized_partial` builds its request separately and already reports `Failed to parse context: <cause>`.

This change relabels the context to `failed to build request`, which is what that step does, and makes `stringify_errors` render with `{:#}` so every returned diagnostic carries its chain. A null context now reports ``failed to build request: while parsing context, found a `null`; JSON `null`s are not allowed in Cedar``. Errors without a chain render exactly as before, so the exact-string assertions for policy-parse, principal-parse, and request-validation diagnostics in `tests/unit/test_authorize_partial.py` still hold. Entity-parse failures were also wrapped in a context and now show their cause after the entities text. `test_is_authorized_with_a_request_that_errors` now asserts the new label and the real cause (the `addPhoto` context is missing its required `photo` attribute), and a new test passes a `null` and a float as `context` with no schema and checks that a cause follows the label. `CHANGELOG.md` gets a `Fixed` entry under `Unreleased`.

## Testing

Run from the repository root with cargo 1.98.0, rustc 1.98.1, maturin 1.15.0, and CPython 3.13.15 on macOS arm64.

```
$ cargo build
   Compiling cedarpy v4.12.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 6.08s
(exit 0)

$ python3 -m venv .venv-fix && . .venv-fix/bin/activate
$ pip install "pytest>=9.0.3" parameterized==0.9.0
$ maturin develop
Installed cedarpy-4.12.0

$ python -m pytest tests/unit/test_authorize.py -q
31 passed, 4 subtests passed in 1.10s

$ python -m pytest -q
8 failed, 267 passed, 16 subtests passed in 1.83s
```

The 8 failures are all in `tests/unit/test_validate.py` and are `FileNotFoundError` on `third_party/cedar/cedar-policy-cli/sample-data/...`: the `cedar` git submodule is not initialized in this checkout. They are unrelated to this change and fail the same way on `main` in the same checkout.

Before and after, on the same request with `schema=None` (reproduced by rebuilding with `main`'s `src/lib.rs`, then with this branch's):

```
context = "null"
  main:   ['failed to parse schema from request']
  branch: ['failed to build request: while parsing context, found a `null`; JSON `null`s are not allowed in Cedar']

context = '{"authenticated": true, "score": 1.5}'
  main:   ['failed to parse schema from request']
  branch: ['failed to build request: data did not match any variant of untagged enum RawCedarValueJson']

principal = 'User::alice'
  main:   ['failed to parse schema from request']
  branch: ['failed to build request: Failed to parse principal as entity Uid: unexpected end of input']

entities = '[{'
  main:   ['failed to parse entities from:\n[{']
  branch: ['failed to parse entities from:\n[{: error during entity deserialization: EOF while parsing an object at line 1 column 2']
```


skuenzli:
Thank you David. I will take a look.

skuenzli:
Thank you for the fix! Much appreciated.

